PR #4, opened 2017-07-03
Interop parser

johandahlberg:
Adding a parser for Interop files. Right now it only reads the error rates per lane and read, but we should be able to extend it to do more later if necessary.


checkQC/parsers/interop_parser.py
```diff
@@ -0,0 +1,53 @@
...
+                return True
+            else:
+                return False
+
+        def __hash__(self):
+            return hash(self.runfolder)
+
+    instance = None
+
+    def __init__(self, runfolder, *args, **kwargs):
```
withrocks: Regarding the singleton, I think this will work as expected. But if a user initializes the object again with other arguments, he has no indication of that he's actually getting that old instance. Perhaps one might want to add an `else raise` after the if?

withrocks (commented):
Looks good as far as I can tell. I've just that one comment about the singleton.

johandahlberg:
@withrocks perhaps you'd like to take a look now? I've made the changes we talked about, and also wrote some tests for the `QCEngine`. On a whole I think it turned our very good after I followed the pattern you suggested.

johanherman (approved):
Always messy to review these kinds of PRs, but I like the factory. :+1:

PR #5, opened 2017-07-04
Draft for real config

johandahlberg:
Here is a draft of what the real config might look like, as you can see all the handlers required are not there yet, and I only wrote it out for the HiSeq2500 v4, but it's mostly to get a feel for what it might look like.

How do you folks want to do with the warning levels? Since the Canea docs do not specify any thing like that, but only give the hard limits, I've left them all as `unknown` here. Does this look reasonably to you?

### DIFF
--- a/config/real_config.yaml
+++ b/config/real_config.yaml
@@ -1,0 +1,53 @@
+
+hiseq2500_v4:
+  50-70:
+    handlers:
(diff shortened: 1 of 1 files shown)


monikaBrandt (approved):
Looks good! I suggest that we change PoolingEvennessHandler to SampleFractionHandler.

PR #6, opened 2017-07-31
Full transfer of qc criteria docs into config

johandahlberg:
I've transferred the QC criteria docs into the config. One thing which we need to discuss is how to interpret the intervals in those documents. Are they inclusive at both ends or not? I would assume that's how they are to be interpreted, and that those which have only a single value will always be at that exact value? This is something that we need some extra code to parse later on.

### DIFF
--- a/config/real_config.yaml
+++ b/config/real_config.yaml
@@ -1,13 +1,15 @@
 
-hiseq2500_v4:
+default_handlers:
+    - name: UndeterminedPercentageHandler
+      warning: unknown
+      error: 10
+    - name: PoolingEvennessHandler
+      error: 50 # Each sample should have  >= x % of the clusters divded by the number of samples in the pool
+      warning: unknown
+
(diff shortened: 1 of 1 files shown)


matrulda (commented):
Looks good! I interpret the read length intervals the same way as you do.
Two minor things:
- I'm thinking that most of the error thresholds should instead be warnings. This is not something that is specified in the QC criteria document, but something that needs to be in place for the production to run smoothly. We could add this later though. 
- It would be nice if error and warning always came in the same order, to prevent from filling in the wrong field.

johandahlberg:
Good point about the ordering.

Regarding the levels for error vs. warning this is just what I copied over from the QC criteria document. I think it would be great if you could have a look at it and make a PR later with the exact values you want it to have for the respective levels.

Feel free to merge this once you think it's alright.

matrulda (approved):
Great!

PR #7, opened 2017-08-08
Introducing RunTypeRecognizer

johandahlberg:
The `RunTypeRecognizer` is a class aiming at recognizing different types of Illumina instruments and run modes, so that the correct quality control config can be applied later.

### DIFF
--- a/checkQC/app.py
+++ b/checkQC/app.py
@@ -1,22 +1,23 @@
 
-import yaml
 import sys
 
 from checkQC.qc_engine import QCEngine
+from checkQC.config import get_config
+from checkQC.run_type_recognizer import RunTypeRecognizer
 
 
 def start():
 
-    with open("config/config.yaml") as stream:
-        config = yaml.load(stream)
+    config = get_config("config/config.yaml")
 
-    machine_type = "hiseq2500_rapid"
-    read_length = '100-120'
-    run_mode = "paired_end"
+    runfolder = "./tests/resources/MiSeqDemo"
 
-    handler_config = config[machine_type][read_length][run_mode]["handlers"]
+    run_type_recognizer = RunTypeRecognizer(config=config, runfolder=runfolder)
+    instrument_type = run_type_recognizer.instrument_type()
+    run_mode = run_type_recognizer.single_or_paired_end()
+    read_length = run_type_recognizer.read_length()
 
-    runfolder = "./tests/resources/150418_SN7001335_0149_AH32CYBCXX"
+    handler_config = config[instrument_type][read_length][run_mode]["handlers"]
 
     qc_engine = QCEngine(runfolder=runfolder, handler_config=handler_config)
     qc_engine.run()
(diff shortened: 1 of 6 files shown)


checkQC/run_type_recognizer.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def single_or_paired_end(self):
+        """
+        Attempts to determine if the run has been run as single or paired end run.
+        :return: "single_end" or "paired_end"
+        :raises: RunModeUnknown
+        """
+        reads = self._run_info["RunInfo"]["Run"]["Reads"]["Read"]
+        nbr_of_reads = len(reads)
+        if nbr_of_reads == 1:
```
matrulda: This will not work when index reads are included. For example, a paired-end run with dual index will list 4 reads.  

I actually don't think we need to know if it's paired-end or single, the QC criteria are the same for both. But it is needed for the mock config.

checkQC/run_type_recognizer.py
```diff
@@ -0,0 +1,82 @@
...
+        """
+        Gathers information on the read length of the run.
+        :return: The read length. If multiple reads delimited by "-"
+        """
+        reads = self._run_info["RunInfo"]["Run"]["Reads"]["Read"]
+
+        read_lengths = []
+        for read in reads:
+            if not read['@IsIndexedRead'] == 'Y':
+                #TODO unsure about the -1 here...
```
matrulda: Yeah, this is confusing, but I think this is correct. To make the readlength correspond with what is written in the documentation we need to subtract one.

matrulda (commented):
This is a great start! 
If we are going to keep the single or paired-end check, I would like it to work with indexed runs. I do however doubt that we need it. I'm OK with it not being fixed if it is removed when the RunTypeRecognizer is adapted for the real_config. 

Note for further development: To make this work with the "real_config", we need to get the reagent kit version from [rR]unParameters.xml.

checkQC/run_type_recognizer.py
johandahlberg: Good point. I'll remove this method for now, and make the necessary changes to the rest of the code to make sure it works. This was an erroneous assumptions on my part.

checkQC/app.py
```diff
@@ -1,22 +1,22 @@
...
 
-    machine_type = "hiseq2500_rapid"
-    read_length = '100-120'
-    run_mode = "paired_end"
+    runfolder = "./tests/resources/MiSeqDemo"
 
-    handler_config = config[machine_type][read_length][run_mode]["handlers"]
+    run_type_recognizer = RunTypeRecognizer(config=config, runfolder=runfolder)
+    instrument_type = run_type_recognizer.instrument_type()
+    read_length = run_type_recognizer.read_length()
```
matrulda, line 17: Ah sorry, discovered a new thing now. 
As I understand it, read_length() will (if paired-end data) return `<readlenght1>-<readlength2>`. This will not work when declaring the handler_config (row 19), right? The read length interval has to be determined before extracting the handler config.

johandahlberg: Yes. You are right, but I think it is something that can be left for later when we integrate the whole thing with the "real" config. There are some things that we need to figure out here, such as how to handle when we've got differing read lengths on read 1 and read 2 etc. If you're ok with it perhaps we can merge now, and then we can fix it later?

PR #10, opened 2017-08-09
Add command line interface to app

johandahlberg:


checkQC/app.py
```diff
@@ -1,27 +1,54 @@
...
+import click
+
 from checkQC.qc_engine import QCEngine
-from checkQC.config import get_config
+from checkQC.config import get_config, get_handler_config
 from checkQC.run_type_recognizer import RunTypeRecognizer
 
 
-def start():
+@click.command("checkQC")
```
withrocks: A minor UX thing: Will you guys not get tired of having to switch case when writing this command?

checkQC/app.py
```diff
@@ -1,27 +1,54 @@
...
+    sys.exit(app.exit_status)
+
+
+class App(object):
 
-    config = get_config("config/config.yaml")
+    def __init__(self, runfolder, config_file=None):
+        self._runfolder = runfolder
+        self._config_file = config_file
+        self.exit_status = 0
```
withrocks, line 34: One might perhaps want to consider throwing well defined exceptions in the App class, and only translate them to exit codes in the CLI. Then the App class would be more usable as a library.

checkQC/app.py
```diff
@@ -1,27 +1,54 @@
...
 
-    runfolder = "./tests/resources/MiSeqDemo"
+    def run(self):
+        config = get_config(self._config_file)
 
-    run_type_recognizer = RunTypeRecognizer(config=config, runfolder=runfolder)
-    instrument_type = run_type_recognizer.instrument_type()
-    read_length = run_type_recognizer.read_length()
+        run_type_recognizer = RunTypeRecognizer(config=config, runfolder=self._runfolder)
```
withrocks, line 39: I like that name :)

checkQC/app.py
```diff
@@ -1,27 +1,54 @@
...
+        instrument_type = run_type_recognizer.instrument_type()
+        reagent_version = run_type_recognizer.reagent_version()
 
-    handler_config = config[instrument_type][read_length]["handlers"]
+        # TODO For now assume symmetric read lengths
+        read_length = int(run_type_recognizer.read_length().split("-")[0])
```
withrocks, line 44: Would it make sense for the recognizer to return the read_length? Then only it knows the format.

checkQC/app.py
```diff
@@ -1,27 +1,54 @@
...
 
-    qc_engine = QCEngine(runfolder=runfolder, handler_config=handler_config)
-    qc_engine.run()
-    sys.exit(qc_engine.exit_status)
+        instrument_and_reagent_type = "_".join([instrument_type, reagent_version])
 
+        handler_config = get_handler_config(config, instrument_and_reagent_type, read_length)
+        qc_engine = QCEngine(runfolder=self._runfolder, handler_config=handler_config)
+        qc_engine.run()
+        self.exit_status = qc_engine.exit_status
```
withrocks, line 51: ... (see above) so perhaps remove this and have a try/except in the CLI.

checkQC/config.py
```diff
@@ -1,8 +1,30 @@
...
+
 import yaml
 
 
 def get_config(config_file):
-    with open(config_file) as stream:
-        config = yaml.load(stream)
-    return config
+    try:
+        if not config_file:
```
withrocks: Very minor: Might want to call it `config_path` or `config_file_name` so it's clear which one is the path and which has the file content.

checkQC/config.py
```diff
@@ -1,8 +1,30 @@
...
+            config_file = resource_filename(Requirement.parse('checkQC'), 'checkQC/default_config/config.yaml')
+            print("No config file specified, using default config from {}.".format(config_file))
+
+        with open(config_file) as stream:
+            config = yaml.load(stream)
+        return config
```
withrocks: I believe you can return from within the with statement. One line of code removed!

checkQC/config.py
```diff
@@ -1,8 +1,30 @@
...
+def get_handler_config(config, instrument_and_reagent_type, read_length):
+    try:
+        handler_config = config[instrument_and_reagent_type][read_length]["handlers"]
+        return handler_config
+    except KeyError as e:
+        print("Could not find a config entry for instrument '{}' "
+              "with read length '{}'. Please check the provided config "
+              "file ".format(instrument_and_reagent_type,
+                             read_length))
+        raise e
```
withrocks: Perhaps want to add that newline if any strange old unix tools need to read this file.

checkQC/run_type_recognizer.py
```diff
@@ -29,6 +29,28 @@ def __init__(self, config, runfolder):
         with open(os.path.join(self._runfolder, "RunInfo.xml")) as f:
             self._run_info = xmltodict.parse(f.read())
 
+        with open(self._find_run_parameters_xml()) as f:
+            self._run_parameters = xmltodict.parse(f.read())
+
+    def _find_run_parameters_xml(self):
+        first_option = os.path.join(self._runfolder, "RunParameters.xml")
```
withrocks, line 36: Looks good, but just a suggestion for a more generic solution (here I assume that any case combination is fine):

```
files = [os.path.join(self._runfolder, f) for f in os.listdir(self._runfolder)
         if os.path.isfile(f) and f.lower() == 'runparameters.xml']
if len(files) == 1:
  return files[0]
else:
  raise ...
```

Side note: In `clarity-ext` we actually have a helper for this, modelled on a very popular built-in in .NET. Equipped with that helper, this style makes much more sense (since the code becomes very clear, methinks):
```
files = [os.path.join(self._runfolder, f) for f in os.listdir(self._runfolder)
         if os.path.isfile(f) and f.lower() == 'runparameters.xml']
return utils.single(files)  # Raises if it's not exactly one.
```

checkQC/run_type_recognizer.py
```diff
@@ -29,6 +29,28 @@ def __init__(self, config, runfolder):
...
+        second_option = os.path.join(self._runfolder, "runParameters.xml")
+        if os.path.isfile(first_option):
+            return first_option
+        elif os.path.isfile(second_option):
+            return second_option
+        else:
+            raise FileNotFoundError("Could not find [R|r]arameters.xml for runfolder {}".format(self._runfolder))
```
withrocks: Should probably be `[R|r]unParameters.xml `

withrocks (approved):
Looks good. Some minor suggestions attached.

withrocks:
Let me know when you want this merged.

checkQC/app.py
johandahlberg: Good point, I'll fix this.

checkQC/app.py
johandahlberg: The reason I want to do it this way is that when the data does not pass the qc criteria a non-zero exit status should be returned from the program, even when no exceptions has been thrown.

checkQC/app.py
johandahlberg: I'll leave this for later as I'm still not sure if we need to handle the case of asymmetric read lengths, e.g. `300-150` which I think are theoretically possible, though not found in our internal docs.

checkQC/config.py
johandahlberg: Good point.

checkQC/config.py
johandahlberg: :+1:

checkQC/run_type_recognizer.py
johandahlberg: Nice pattern. Will remember it for later, though for this case I find the code as is easier to understand.

checkQC/app.py
withrocks: Ahh, I meant that you would throw a particular exception instead of setting the exit code, but it's a minor thing anyways.

johandahlberg: Ok. The I get it.

PR #12, opened 2017-08-17
Cluster pf handler and handler test implemented

monikaBrandt:


checkQC/handlers/cluster_pf_handler.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def check_qc(self):
+
+        for lane_dict in self.conversion_results:
+            lane_nbr = lane_dict["LaneNumber"]
+            lane_pf = lane_dict["TotalClustersPF"]
+
+            if self.qc_config["error"] != "unknown" and lane_pf <= float(self.qc_config["error"])*pow(10, 6):
+                yield QCErrorFatal("Clusters PF was to low on lane {}, it was: {}".format(lane_nbr, lane_pf))
+            elif lane_pf <= float(self.qc_config["warning"])*pow(10, 6):
```
johandahlberg: I'd suggest checking for unknown here as well, for consistency for the user.

tests/handlers/test_cluster_pf_handler.py
```diff
@@ -0,0 +1,60 @@
+import unittest
+
+from checkQC.handlers.cluster_pf_handler import ClusterPFHandler
+from checkQC.handlers.qc_handler import QCErrorFatal, QCErrorWarning
+
+from tests.test_utils import get_stats_json
+
+class TestClusterPFHandler(unittest.TestCase):
```
johandahlberg: Since both `set_qc_config` and `map_errors_and_warnings_to_class_names` are used in the other handler class as well, perhaps you could look into setting this up with a `HandlerUnitTest` super class and let the other ones inherit from that? That would leave less code to write later.

johandahlberg (changes requested):
This looks nice. I this we should make the changes I suggest regarding `unknown`, but the suggestion about the test class is something you could do if you want to.

Great work!

checkQC/handlers/cluster_pf_handler.py
monikaBrandt: Do you mean in the elif-statement? I want it to throw the same warning no matter what the limit for fatal error is. Can you please specify a bit how you think :)

johandahlberg: Here is what I meant:

```
            if self.qc_config["error"] != "unknown" and lane_pf <= float(self.qc_config["error"])*pow(10, 6):
                yield QCErrorFatal("Clusters PF was to low on lane {}, it was: {}".format(lane_nbr, lane_pf))
            elif self.qc_config["warning"] != "unknown" and lane_pf <= float(self.qc_config["warning"])*pow(10, 6):
                yield QCErrorWarning("Cluster PF was to low on lane {}, it was: {}".format(lane_nbr, lane_pf))
            else:
                continue
```

johandahlberg:
Good catch @monikaBrandt. It's fixed now.

PR #13, opened 2017-09-05
Configs will now allow for both interval and exact match for read lengths

johandahlberg:
Correct handler configurations can now be fetched for both exact matches and interval matches for read lenghts.


johandahlberg:
I also added some tests for the stats json parser, since I noticed I'd not written any previously.

monikaBrandt (approved):
Looks great! Good job!

PR #15, opened 2017-09-08
Sorting error/warning logs from by the lane id

johandahlberg:
This should make sure that within each handler errors/warnings are emitted in a order specified by the user, for our case I'm guessing this will tend to be the lane order.

I have switched to using proper logging, instead of using if statements, to allow this to be tested more easily, which should also make it easier for us to use this code in a library at a later date if we would like to do so.

### DIFF
--- a/checkQC/handlers/qc_handler.py
+++ b/checkQC/handlers/qc_handler.py
@@ -1,4 +1,7 @@
 
+import logging
+
+log = logging.getLogger()
 
 class Subscriber(object):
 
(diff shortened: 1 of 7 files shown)


checkQC/handlers/qc_handler.py
```diff
@@ -19,33 +22,31 @@ def send(self, value):
...
-        return "Warning: {}".format(self.message)
+        return "QC warning: {}".format(self.message)
 
     def __repr__(self):
         return self.__str__()
 
 
 class QCHandler(Subscriber):
 
-    handlers = []
```
johandahlberg, line 45: Just like to add that I removed this code since it is not used, and I appear to have forgotten to remove it previously.

tests/handlers/test_qc_handler.py
```diff
@@ -0,0 +1,35 @@
...
+
+class TestQCHandler(unittest.TestCase):
+
+    random_qc_errors_and_warning = [QCErrorFatal("1", 1), QCErrorWarning("2", 2),
+                                    QCErrorWarning("3", 3), QCErrorFatal("4", 4)]
+
+    class MockQCHandler(QCHandler):
+        # This is a method to mock returning results from the handler
+        def check_qc(self):
+            return TestQCHandler.random_qc_errors_and_warning
```
matrulda: Maybe I'm missing something, but wouldn't it be better to add an unordered list to test that it actually returns the warnings/errors ordered?

matrulda (commented):
Looks good! I had one comment you can take a look at.

johandahlberg:
You had a very good point about the test. I've fixed that now. Feel free to merge when you're ready.

PR #18, opened 2017-09-20
Q30 handler

matrulda:
checkQC can now check percent Q30.

Also discovered that some tests were "cheating" by checking if true instead of equal.


johandahlberg (approved):
This looks great. I'm ready to merge when you are. One general point that I'd like to raise is that right now we are inconsistent in if cut-offs are inclusive or exclusive (i.e. `>=` or `>`), which is something that I think we should harmonize. Which do you think makes the most sense?

PR #19, opened 2017-09-22
Reading default config and have instrument specific pickup of reagent version

johandahlberg:
This PR does to things:

1) It ensures that the default handlers are loaded, unless a handler with the same name has already been defined for that instrument.
2) It allows getting the reagent version to be specified on a per instrument type level. This means that to introduce new instruments one needs to actually add code, which is not optimal, but I think that for now this is the most pragmatic way of solving the problem.

@monikaBrandt Check the `TODO` tags for where to place the code to check if a run is rapid/highoutput.


monikaBrandt (approved):
Looks good. Nice job!

PR #22, opened 2017-10-02
Fix integration test

johandahlberg:
This fixes a number of minor issues I found when testing with the new dataset:

 - I saw that the samplesheet was called `SampleSheet.out` instead of `SampleSheet.csv`
 - I replaced the old dummy config with the real config new, with the exception that I commented out the sample fraction part (we'll need to add that back later)
 - Removed the unused simple file parser, which I just wrote as a proof-of-concept at the start of the project
 - The undetermined percentage handler should be able to handle the scenario where there is no "Undetermined" for a lane, which there isn't e.g. when no indexes has been specified on that lane
 - The Interop parser should send a tupple, not a index, which it wan't doing at this point
 - I change from `config_file` to `config` in the cli, since I kept writing it wrong I assume others will do the same
 - Change the integration tests to use the new data


monikaBrandt (approved):
Great job!

PR #24, opened 2017-10-05
Reads per sample handler

matrulda:
Handler for reads per sample criteria. I changed the name (originally called sample fraction), since I think it makes more sense.


monikaBrandt:
Everything looks great!

PR #27, opened 2017-10-31
Validate configurations and give error on missing error rate

johandahlberg:
This introduces two thing:
 - The possibility for handlers to validate their own configuration prior to running. By default they will check that they have a `warning` and `error` field, but by implementing the `custom_configuration_validation` method any type of config can be checked.
 - When the error rate is zero, it is likely that it means that no PhiX was added to that lane. The default will be to throw an error and exit with a non-zero exit status if this happens, but it can be configured by setting the `allow_missing_error_rate` to `True` in the config.

### DIFF
--- a/checkQC/config.py
+++ b/checkQC/config.py
@@ -7,6 +7,10 @@ import yaml
 log = logging.getLogger("")
 
 
+class ConfigurationError(Exception):
+    pass
+
+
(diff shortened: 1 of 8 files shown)


monikaBrandt (approved):
All is well as far as I can see. Good job!

PR #31, opened 2017-11-08
Attempting to fix the behavior of codecov

johandahlberg:


johandahlberg:
I'm going to self merge this as it is only a configuration change.

PR #38, opened 2017-11-17
Improved information in log output and README

johandahlberg:
This PR adds more information into the output, both in the normal logging and in the `--json` mode. The main thing is that it will include the thresholds set for the handlers into the output, but it will also tell the user which run type it has determined for the runfolder.


README.md
```diff
@@ -49,6 +49,154 @@ When CheckQC starts and no path to the config file is specified it will give you
...
+WARNING  QC warning: Cluster PF was to low on lane 1, it was: 117.93 M
+WARNING  QC warning: Cluster PF was to low on lane 7, it was: 122.26 M
+WARNING  QC warning: Cluster PF was to low on lane 8, it was: 177.02 M
+ERROR    Fatal QC error: Number of reads for sample Sample_pq-27 was too low on lane 7, it was: 6.893 M
+ERROR    Fatal QC error: Number of reads for sample Sample_pq-28 was too low on lane 7, it was: 7.104 M
+INFO     Finished with fatal qc errors and will exit with non-zero exit status.
+```
+
+The program will summarize the type of run it has identified and output any warnings and/or errors in finds.
+If no errors were found the CheckQC will output a non-zero exit status. This means it can easily be used to
```
b97pla: shouldn't it be a zero exit status?

README.md
```diff
@@ -49,6 +49,154 @@ When CheckQC starts and no path to the config file is specified it will give you
...
+decide if a further steps should run or not, e.g. in a workflow.
+
+In addition to the normal output CheckQC has a json mode, enabled by adding `--json` to the commandline.
+This output the results normally shown in the log as json on `stdout`, so that this can either be written to
+a file, or redirected to other programs which can parse the data further. In this example we use the python
+json tool to pretty print the json output:
```
b97pla: wouldn't it be good if the log output could either be suppressed or redirected to e.g. stderr or a file? perhaps it will mess up piping and parsing to have the json output together with the log output?

b97pla (approved):
Just a couple of suggestions but I'm happy. Update or merge at will 👍

README.md
johandahlberg: Good point. I'll make this clearer.

README.md
johandahlberg: Log output is already on `stderr` (as it should be :wink:)

b97pla:
Great!

PR #40, opened 2017-11-20
Improve exceptions handling

johandahlberg:
Introducing a number of specific `CheckQCException`s, so that we can handle those gracefully without printing a stacktrace.

### DIFF
--- a/checkQC/run_type_recognizer.py
+++ b/checkQC/run_type_recognizer.py
@@ -109,6 +109,7 @@ class RunTypeRecognizer(object):
             run_info_path = os.path.join(self._runfolder, "RunInfo.xml")
             if not os.path.exists(run_info_path):
                 log.error("Could not find a RunInfo.xml in {}. Are you sure this is a runfolder?".format(run_info_path))
+                raise FileNotFoundError("Could not find {}".format(run_info_path))
             with open(run_info_path) as f:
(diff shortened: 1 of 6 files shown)


checkQC/run_type_recognizer.py
```diff
@@ -113,11 +105,20 @@ def __init__(self, config, runfolder):
         """
         self._config = config
         self._runfolder = runfolder
-        with open(os.path.join(self._runfolder, "RunInfo.xml")) as f:
-            self.run_info = xmltodict.parse(f.read())
+        try:
+            run_info_path = os.path.join(self._runfolder, "RunInfo.xml")
+            if not os.path.exists(run_info_path):
+                log.error("Could not find a RunInfo.xml in {}. Are you sure this is a runfolder?".format(run_info_path))
```
withrocks, line 111: Shouldn't this also raise an error?

withrocks (approved):

checkQC/run_type_recognizer.py
johandahlberg: You are right, that might be better. Right now it is raised by the `open` statement below instead, but it might be better to do it explicitly instead.

withrocks: :+1:

PR #46, opened 2018-01-29
Interop can now be installed via PyPI

johandahlberg:
The Interop package can now be installed via PyPI, which means that it can be automatically installed via the setup file, instead of being manually installed prior to installing CheckQC.


johandahlberg:
I fixed a minor problem with the paper missing a `@` in one of the references.

b97pla:
:+1:

Issue #47, opened 2018-01-31
Path to bcl2fastq output in config


matrulda:
CheckQC assumes that output from bcl2fastq is located in a sub directory called `Unaligned`, but output is by default written to `<RUNFOLDER>/Data/Intensities/BaseCalls`. It would be nice if the location of the bcl2fastq output could be specified in the config.

PR #48, opened 2018-02-01
Configurable bcl2fastq output

johandahlberg:
This relates to what has been reported in:
 - https://github.com/openjournals/joss-reviews/issues/556
 - #47 

Basically this code makes it possible to configure where bcl2fastq has written its output, and switches the default to match the standard output location of bcl2fastq.

The way to configure it from now on is to set it in the newly introduced parser config section in the config, e.g:

```
parser_configurations:
  StatsJsonParser:
    bcl2fastq_output_path: Data/Intensities/BaseCalls
```


matrulda (approved):
Great!

PR #50, opened 2018-02-02
Displayed threshold is now divided by the number of samples in the lane

### LINKED ISSUE #49: Reads per sample handler always report threshold based on one sample per lane
https://github.com/Molmed/checkQC/blob/35b30acfbf5453283eda90ae245bf9e9e7edb8df/checkQC/handlers/reads_per_sample_handler.py#L48

I noticed that the threshold reported is not based on the number of samples in the lane. This is confusing. It would make more sense to show the value that was evaluated.

matrulda:
Fixes #49

### DIFF
--- a/checkQC/handlers/reads_per_sample_handler.py
+++ b/checkQC/handlers/reads_per_sample_handler.py
@@ -39,20 +39,25 @@ class ReadsPerSampleHandler(QCHandler):
                 sample_id = sample_id_info["SampleId"]
                 sample_total_reads = sample_id_info["NumberReads"] / pow(10, 6)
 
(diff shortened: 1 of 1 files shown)


johandahlberg:
Great. Can I suggest that you break out the warning and error levels to their own variable, since now there is some unnecessary code duplication going on? That would make it easier to change if we need to change the business logic here in the future.

johandahlberg (changes requested):
See my other comment.

matrulda:
Great suggestion. Have a look now @johandahlberg . The UNKNOWN check is done twice, but this was the most "elegant" solution I could come up with.

johandahlberg (approved):

PR #54, opened 2018-02-15
Read length equal to number of cycles

### LINKED ISSUE #53: Error when using lower bound in read length interval
I found a bug when processing a `hiseq2500_rapidhighoutput_v4` with read length `50`. 

ERROR: 
```
INFO     ------------------------
INFO     Starting checkQC (1.1.2)
INFO     ------------------------
INFO     Runfolder is: /data/biotank7/runfolders/180124_D00457_0238_BCB2B1ANXX
INFO     No config file specified, using default config from /opt/checkqc/lib/python3.6/site-packages/checkQC/default_config/config.yaml.
ERROR    Could not find a config entry for instrument 'hiseq2500_rapidhighoutput_v4' with read length '50'. Please check the provided config file 
Traceback (most recent call last):
  File "/opt/checkqc/bin/checkqc", line 11, in <module>
    sys.exit(start())
  File "/opt/checkqc/lib/python3.6/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/opt/checkqc/lib/python3.6/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/opt/checkqc/lib/python3.6/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/opt/checkqc/lib/python3.6/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/opt/checkqc/lib/python3.6/site-packages/checkQC/app.py", line 36, in start
    app.run()
  File "/opt/checkqc/lib/python3.6/site-packages/checkQC/app.py", line 66, in run
    reports = self.configure_and_run()
  File "/opt/checkqc/lib/python3.6/site-packages/checkQC/app.py", line 55, in configure_and_run
    handler_config = config.get_handler_config(instrument_and_reagent_version, read_length)
  File "/opt/checkqc/lib/python3.6/site-packages/checkQC/config.py", line 90, in get_handler_config
    raise e
  File "/opt/checkqc/lib/python3.6/site-packages/checkQC/config.py", line 82, in get_handler_config
    handler_config = self._get_matching_handler(instrument_and_reagent_type, read_length)
  File "/opt/checkqc/lib/python3.6/site-packages/checkQC/config.py", line 64, in _get_matching_handler
    raise KeyError
KeyError
```

Config: 
```
[...]
hiseq2500_rapidhighoutput_v4:
  50-70:
    handlers:
      - name: ClusterPFHandler
        warning: 180 # Millons of clusters
        error: unknown
      - name: Q30Handler
        warning: 80 # Give percentage for reads greater than Q30
        error: unknown # Give percentage for reads greater than Q30
      - name: ErrorRateHandler
        allow_missing_error_rate: False
        warning: 1.5
        error: unknown
      - name: ReadsPerSampleHandler
        warning: 90 # 50 % of threshold for clusters pass filter
        error: unknown
  100-110:
[...]
```

I think the problem is that we are defining the read length as number of cycles - 1: 
https://github.com/Molmed/checkQC/blob/8911c2844b817bd85a9dd4c7d788b8d835ff8dd3/checkQC/run_type_recognizer.py#L221

and then doing an exclusive check against the lower bound: 
https://github.com/Molmed/checkQC/blob/8911c2844b817bd85a9dd4c7d788b8d835ff8dd3/checkQC/config.py#L94

So in my case: 51 cycles were run for read1, it is adjusted to read length 50. Then we check if 50 < 50, which is not true. 

This could be fixed by not defining read length as number of cycles minus one or by doing an inclusive check against the lower bound. What do you think is the best, @johandahlberg @monikaBrandt ?

matrulda:
- The read length is now equal to the number of "read cycles". 
 - Read length interval is now inclusive on both sides. I find this less confusing. 
 - Moved config info to docs. 

Fixes #53


checkQC/config.py
```diff
@@ -91,12 +92,15 @@ def _get_matching_handler(self, instrument_and_reagent_type, read_length):
                 split_read_length = config_read_length.split("-")
                 low_break = int(split_read_length[0])
                 high_break = int(split_read_length[1])
-                if low_break < int(read_length) <= high_break:
+                if low_break <= int(read_length) <= high_break:
                     return self._config[instrument_and_reagent_type][config_read_length]["handlers"]
             else:
                 if int(read_length) == int(config_read_length):
                     return self._config[instrument_and_reagent_type][int(config_read_length)]["handlers"]
-        raise KeyError
```
johandahlberg, line 99: I think that the logging that we were previously doing in the `get_handler_config` method should go here, so that this error is communicated to the user. Right now I don't think any indication of this will show up to the user.

docs/index.rst
```diff
@@ -199,6 +199,25 @@ In this example we use the python json tool to pretty print the json output:
       }
   }
 
+Configuration file
```
johandahlberg, line 202: This is great. Perhaps you could write something here about that the location of the default config file can be seen when running the program, so that people who wonder where it is located can find it?

johandahlberg (changes requested):
Great work. I have two general comments, in addition to the inline comments.
 - Add `raises` info to the doc string in `_get_matching_handler` to indicate that it can raise a `ConfigEntryMissing` exception.
 - I do not think what we should change the way the ranges are defined. There are two reasons for that. Firstly our current internal qc criteria documents follow this definition, so keeping it makes it easier for us to keep up to date with that. Secondly, and not as important, I find that it makes the intervals less aesthetically pleasing.

monikaBrandt:
Regarding the definition of our read length intervals I think we should do the following:
-Define intervals as upper and lower inclusive to make it less confusing for the common user.  I interpret the intervals in INS-00123 as upper and lower inclusive.
-Create a default config and add to the CheckQC repo where intervals looks nice, i.e. 50-70.
-Create our own SNP&SEQ-config where we add one base to upper and lower bound of the interval because that is how we define read length. 
-Compare intervals to a non-modified value, i.e. the value in runInfo.xml.

I think we should make the software as easy to understand as possible for the common user and keep all platform specific requirements in our own config. If we do as suggested above, we don't need to make any modifications of the value in runInfo.xml before comparing it to the interval. I think all modifications should be made in the configuration file.

b97pla:
FWIW: I agree with @monikaBrandt in that the software should be consistent and use the number of cycles as keys rather than second-guessing how the instruments will behave and what the user wants. @johandahlberg has a good point about the correspondence with the internal documentation though, perhaps we should update that document instead to correspond to number of cycles (or just mention the actual number of cycles in that document?).

Alternatively, is there any field in e.g. RunInfo.xml or runParameters.xml that actually gives a different number as read length (e.g. 150 for 151 cycles)?

johandahlberg:
Ok. :+1: Lets stay with this implementation then. :smile:

matrulda:
I think the best would be to change our internal documentation. I'll leave the config as it is then? 
@johandahlberg Will fix the docs and the exception issue.

checkQC/config.py
```diff
@@ -97,10 +98,7 @@ def _get_matching_handler(self, instrument_and_reagent_type, read_length):
             else:
                 if int(read_length) == int(config_read_length):
                     return self._config[instrument_and_reagent_type][int(config_read_length)]["handlers"]
-        raise ConfigEntryMissing("Could not find a config entry for instrument '{}' "
-                  "with read length '{}'. Please check the provided config "
-                  "file ".format(instrument_and_reagent_type,
-                                 read_length))
+        raise ConfigEntryMissing
```
johandahlberg: No need to remove the message here. It's great to have when debugging.

matrulda:
@johandahlberg See new commits.

checkQC/default_config/config.yaml
```diff
@@ -1,21 +1,6 @@
...
-#        warning: 180 # Millons of clusters
-#        error: unknown
-#
-#    are specific to that partiular handler, but in general any value can be substituted with "unknown", in which case
-#    this will not be evaluated.
-#
-#  - Handlers specified under "default_handlers" will be run regardless of instrument type. For all other cases
-#    it is possible to specify handlers per instrument and read length interval.
-#
+# For information about config usage, see http://checkqc.readthedocs.io/
```
johandahlberg: Can we make this link to the exact part in the docs? I think the headlines have anchors which should make this possible.

johandahlberg (commented):
Some tiny things left to do, but then I think this is ready for merge.

checkQC/default_config/config.yaml
johandahlberg: I checked now. The link should be: http://checkqc.readthedocs.io/en/latest/#configuration-file

checkQC/config.py
matrulda: Ah ok, I'll restore it.

checkQC/default_config/config.yaml
matrulda: Great, will add it.

matrulda:
Would you consider these changes not backwards-compatible? @monikaBrandt @johandahlberg I'm about to update the version number. 

I'm leaning against major update since read length has been re-defined.

johandahlberg:
I would say minor, since the API has has not been changed.

matrulda:
That's true. Minor it is.

johandahlberg (approved):

PR #58, opened 2018-06-13
Feature/avoid forking from service

b97pla:
This PR removes the `ProcessPoolExecutor` functionality from the web service, basically reducing it to running things in a blocking fashion. The problem with the ProcessPoolExecutor was that it spawned a lot of threads that were left running and wouldn't terminate properly. The most straightforward way to fix it was, for now, to remove the threading. 

Obviously, there are more elegant ways to solve this so feel free to step up 😄 

In addition, a Dockerfile for running the service in a container has been added.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,19 @@
+# Use an official Python runtime as a parent image
+FROM python:3.6-slim
+
+# Set the working directory to /app
+WORKDIR /app
+
+# Copy the current directory contents into the container at /app
+COPY . /app
+
+# Install any needed packages specified in requirements.txt and deploy app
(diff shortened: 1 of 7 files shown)


johandahlberg:
Looks good! Perhaps you could add the information about running in a docker container to the docs instead? I think it makes sense to try to keep the README somewhat "minimal" and add additional information in the docs. Does that sound reasonable to you?

b97pla:
ah, makes sense, I didn't think of that. I'll move the instructions. Do you want to keep the `index.rst` file to mirror the README or can I add them there?

johandahlberg:
Either there, or under the development section. Take whichever one you think makes the most sense.

b97pla:
ok, it went into the `index.rst` file

PR #59, opened 2018-06-13
Automatically push tags to PyPI

johandahlberg:
This should make all releases (i.e. tags) be pushed automatically to PyPI. Nice to take one more step out of the release process.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,18 @@
 language: python
 python:
-  - "3.6"
-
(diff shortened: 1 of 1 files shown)


b97pla:
I don't have much input on this but I'll merge if you think it's ready..

johandahlberg:
Go ahead. Unfortunately one these things that are difficult to test without triggering the Travis event...

PR #60, opened 2018-06-13
Add provider for deploy

johandahlberg:
I obviously missed to add a provider for the deploy. :unamused: However now I did learn of `travis lint .travis.yml `, which I will try to use in the future to avoid this type of mishaps...


b97pla:
alright, let's have another go

PR #67, opened 2018-09-04
Support for iSeq, and improved org for testing

johandahlberg:
This adds support for the Illumina iSeq, and also makes some changes to the way that the RuntypeRecognizer works by introducing a separate class for reading the files in the runfolder. This makes this easier to mock  in the tests, which makes it easier to test that the correct  instrument is returned 
 depending on which type of runfolder is passed  in.

The QC criteria that I have added are completely arbitrary, and need to be update.

### DIFF
--- a/checkQC/run_type_recognizer.py
+++ b/checkQC/run_type_recognizer.py
@@ -1,9 +1,8 @@
 
-import os
 import logging
-import xmltodict
 
 from checkQC.exceptions import *
+from checkQC.runfolder_reader import RunfolderReader
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


checkQC/run_type_recognizer.py
```diff
@@ -57,6 +56,15 @@ def reagent_version(runtype_recognizer):
...
 
+class ISeq(IlluminaInstrument):
+
+    @staticmethod
+    def name():
+        return "iseq"
+
+    @staticmethod
+    def reagent_version(runtype_recognizer):
+        return "v1"
```
b97pla, line 67: does this have to be hard-coded or can it be extracted from `runParameters.xml` instead?

b97pla (approved):
:+1: looks good to me, just had a question

checkQC/run_type_recognizer.py
johandahlberg: That is a very good question. Since for now there is (as far as I know) only one type of reagent kit I think that this might be ok for now?

matrulda: I think this is fine for now.

matrulda (approved):
Good job! Ready to merge.

johandahlberg:
Go ahead and merge @MatildaAslin

PR #68, opened 2018-09-07
Improved error handling

johandahlberg:
This should improve the error handling of CheckQC in general, and for the webservice in particular. Many errors which previously caused timeouts should now fail fast instead.

### DIFF
--- a/checkQC/app.py
+++ b/checkQC/app.py
@@ -62,8 +62,8 @@ class App(object):
         parser_configurations = config.get("parser_configurations", None)
 
         if not Path(self._runfolder).is_dir():
-            raise RunfolderNotFoundError("Could not find runfolder: {} are you "
-                                         "sure the path is correct".format(self._runfolder))
+            raise RunfolderNotFoundError("Could not find runfolder: {}. Are you "
+                                         "sure the path is correct?".format(self._runfolder))
 
         run_type_recognizer = RunTypeRecognizer(runfolder=self._runfolder)
         instrument_and_reagent_version = run_type_recognizer.instrument_and_reagent_version()
--- a/checkQC/app.py
+++ b/checkQC/app.py
@@ -2,6 +2,7 @@
 import sys
 import json
 import logging
+from pathlib import Path
 
 import click
 
@@ -9,7 +10,7 @@ from checkQC.qc_engine import QCEngine
 from checkQC.config import ConfigFactory
 from checkQC.run_type_recognizer import RunTypeRecognizer
 from checkQC.run_type_summarizer import RunTypeSummarizer
-from checkQC.exceptions import CheckQCException
+from checkQC.exceptions import CheckQCException, RunfolderNotFoundError
 from checkQC import __version__ as checkqc_version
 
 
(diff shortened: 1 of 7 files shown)


checkQC/app.py
```diff
@@ -57,29 +58,30 @@ def configure_and_run(self):
...
-            return reports
-        except CheckQCException as e:
-            log.error(e)
-            self.exit_status = 1
+        config = ConfigFactory.from_config_path(self._config_file)
+        parser_configurations = config.get("parser_configurations", None)
+
+        if not Path(self._runfolder).is_dir():
+            raise RunfolderNotFoundError("Could not find runfolder: {} are you "
+                                         "sure the path is correct".format(self._runfolder))
```
mariya: Not sure how nitpicky we need to get about grammar in error messages but if you're in a nitpicky mood... capitalize "are" and add a question mark.

checkQC/app.py
```diff
@@ -92,14 +94,19 @@ def run(self):
...
+            reports = self.configure_and_run()
+            if self.exit_status == 0:
+                log.info("Finished without finding any fatal qc errors.")
+            else:
+                log.error("Finished with fatal qc errors and will exit with non-zero exit status.")
+
+            if self._json_mode:
+                print(json.dumps(reports))
+
+        except CheckQCException as e:
```
mariya, line 107: Do we know of any more specific errors that tend to occur here, that we may want to catch? Fine as is though.

checkQC/config.py
```diff
@@ -87,21 +87,29 @@ def _get_matching_handler(self, instrument_and_reagent_type, read_length):
...
+                if "-" in config_read_length:
+                    split_read_length = config_read_length.split("-")
+                    low_break = int(split_read_length[0])
+                    high_break = int(split_read_length[1])
+                    if low_break <= int(read_length) <= high_break:
+                        return self._config[instrument_and_reagent_type][config_read_length]["handlers"]
+                else:
+                    if int(read_length) == int(config_read_length):
+                        return self._config[instrument_and_reagent_type][int(config_read_length)]["handlers"]
+            raise ConfigEntryMissing("Could not find a config entry matching read length '{}' on "
```
mariya, line 104: Duplicated on line 109... refactor into a helper function?

checkQC/web_app.py
```diff
@@ -31,10 +32,22 @@ def _run_check_qc(monitor_path, qc_config_file, runfolder):
...
+        self.set_status(status_code=status_code)
+        self.finish({"reason": reason})
+
+    def get(self, runfolder):
+        try:
+            reports = self._run_check_qc(self.monitor_path, self.qc_config_file, runfolder)
+            self.set_header("Content-Type", "application/json")
+            self.write(reports)
+        except RunfolderNotFoundError:
+            self._write_error(status_code=404, reason="could not find requested runfolder.")
```
mariya: Capitalize "could" for consistency.

tests/resources/poor_config.yaml
```diff
@@ -0,0 +1,32 @@
+
+# NOTE THIS CONFIG IS HERE TO BE USED TO NOT HAVE THE REQUIRED KEYS IN THE TESTS!
```
mariya: Could "incomplete_config.yml" be a clearer name? Suggestions for comment text that could clarify things a bit: "This config is a partial copy of checkQC/default_config/config.yaml, with required keys removed in order to trigger errors. For testing only."

mariya (changes requested):
Looks really good, always better with clearer error messages! Just a bit of feedback re. names and texts, @withrocks is probably a better bet at this point for feedback re. the actual functionality :)

checkQC/app.py
johandahlberg: :+1:

checkQC/app.py
johandahlberg: In this case we want to catch all "handled" errors since we just log them to the console any way for the command line app.

checkQC/config.py
johandahlberg: They have slightly different messages, so I'm not sure that would go very much. I do notice that I need to turn the argument around here, though.

checkQC/web_app.py
johandahlberg: :+1:

tests/resources/poor_config.yaml
johandahlberg: Good point. I'll fix that.

johandahlberg:
Thanks! I think I have addressed your comments now, so if you think that it looks good you can go ahead and merge.

PR #70, opened 2018-09-21
Check indexes and let phix level decide undetermined threshold

johandahlberg:
This is a rather large PR which setups a proof-of-concept prototype for checking unidentified indexes with checkqc. Internally at SNP&SEQ it relates to issue: https://snpseq.atlassian.net/browse/BIO-1341

I'm pushing this here now because I'd like some feedback at this stage before finalizing things. I have still to write some tests for this, but I'l like to know that I'm on the right track before moving on. Perhaps @MatildaAslin and @b97pla would like to take a look at it and say what you think.


checkQC/default_config/config.yaml
```diff
@@ -60,6 +62,8 @@ hiseq2500_rapidhighoutput_v4:
       - name: ReadsPerSampleHandler
         warning: unknown
         error: 90 # 50 % of threshold for clusters pass filter
+      - name: UnidentifiedIndexHandler
+        significance_threshold: 0.001 # % of reads
```
b97pla: this should say fraction rather than %, right?

checkQC/handlers/unidentified_index_handler.py
```diff
@@ -0,0 +1,131 @@
...
+    def check_reverse_complement_index(self, tag, lane, samplesheet_dict):
+        hit = self.index_in_samplesheet(samplesheet_dict, self.reverse_complement(tag))
+        if hit:
+            msg = 'We found a possible match for the reverse complement of tag: {}, on: {}'.format(tag, hit)
+            yield QCErrorWarning(msg, ordering=lane, data={'lane': lane, 'msg': msg})
+
+    def check_if_index_in_other_lane(self, tag, lane, samplesheet_dict):
+        hit = self.index_in_samplesheet(samplesheet_dict, tag)
+        if hit:
+            msg = 'We found a possible match for the tag: {}, on another lane: {}'.format(tag, hit)
```
b97pla: wouldn't it be useful if we got a list of hits in case the index occurs on multiple lanes?

checkQC/handlers/unidentified_index_handler.py
```diff
@@ -0,0 +1,131 @@
...
+        #     Haven't really figured out how that would work just yet though.
+        #     /JD 2018-09-21
+        return float(index['count']) / nbr_of_reads_on_lane > self.qc_config['significance_threshold']
+
+    def reverse_complement(self, seq):
+        conv_table = {'A': 'T', 'T': 'A', 'G': 'C', 'C': 'G', 'N': 'N', '+': '+'}
+        rev_seq = ''
+        for c in seq:
+            rev_seq += conv_table[c]
+        return rev_seq
```
b97pla: you're never actually reversing the sequence, right? something like `return rev_seq[::-1]`

b97pla (commented):
Looks good, I had a few comments. I'm not entirely familiar with the design of the framework though so I probably wouldn't have spotted any problems with the setup of handlers etc.

checkQC/handlers/unidentified_index_handler.py
johandahlberg: Good point.

checkQC/handlers/unidentified_index_handler.py
johandahlberg: Hmmm... Is that what we want though? Depends on the way the indexes are specified, right? I always have a hard time keeping this straight in my head.

b97pla: I'd say that's what we want but someone might want to second that. In any case, either the method should be renamed to `complement` or the sequence should also be reversed - otherwise it's not the reverse complement 😄

johandahlberg: I think what I did was just to model this after the current Sisyphus code: https://github.com/Molmed/sisyphus/blob/71705b2b0bc091143d5c5ba5c65fef3cff25f0cb/lib/Molmed/Sisyphus/IndexCheck.pm#L282 So right now it just mirrors that functionality and name but I'm good changing it if we are sure that make sense.

b97pla: ok, we want to do both. Test the complement sequence and the reverse complement. So two separate methods with appropriate naming.

johandahlberg: Now there is both. :smile:

b97pla: :+1:

checkQC/default_config/config.yaml
```diff
@@ -8,11 +8,15 @@ parser_configurations:
...
     bcl2fastq_output_path: Data/Intensities/BaseCalls
+  SamplesheetParser:
+    samplesheet_name: SampleSheet.csv
 
 default_handlers:
     - name: UndeterminedPercentageHandler
       warning: unknown
       error: 10
+    - name: UnidentifiedIndexHandler
+      significance_threshold: 0.001 # fraction of reads
```
matrulda: Shouldn't this be 0.01? If we're using the same fraction as in Sisyphus, which is based on QC criteria documents used internally at SNP&SEQ.

checkQC/handlers/unidentified_index_handler.py
matrulda: Great, maybe you can rename `rev_seq` to `comp_seq` or similar to avoid confusion?

matrulda (commented):
Great work! Making the significance threshold dynamic is tricky. A suggestion would be: `1/10*(number of samples in lane)`, but it is ofc arbitrary.

checkQC/handlers/unidentified_index_handler.py
```diff
@@ -0,0 +1,136 @@
...
+        for rule in rules:
+            yield from rule(tag, lane, samplesheet_dict)
+
+    def check_swapped_dual_index(self, tag, lane, samplesheet_dict):
+        if '+' in tag:
+            split_tag = tag.split('+')
+            swapped_tag = '{}+{}'.format(split_tag[1], split_tag[0])
+            hits = self.index_in_samplesheet(samplesheet_dict, swapped_tag)
+            for hit in hits:
+                msg = 'It appears that maybe the dual index tag: {} was swapper. There was a hit for' \
```
matrulda: It would be really useful to include index percentage (that is, how much of the lane does this hit make up for) to the messages.  It makes quite a difference if a hit make up 50% or 1 % of the lane.

checkQC/default_config/config.yaml
johandahlberg: Yeah. This is just a test level which I added to make sure I got something on the folders I tested on.

checkQC/handlers/unidentified_index_handler.py
johandahlberg: :+1:

checkQC/handlers/undetermined_percentage_handler.py
```diff
@@ -45,18 +62,27 @@ def check_qc(self):
...
-                                         "to high on lane {}, it was: {:.2f}%".format(lane_nbr, percentage_undetermined),
+                                         "to high on lane {}, it was: {:.2f}%".format(lane_nbr,
+                                                                                      percentage_undetermined),
                                          ordering=lane_nbr,
-                                         data={"lane": lane_nbr, "percentage_undetermined": percentage_undetermined,
-                                               "threshold": self.warning()})
+                                         data={"lane": lane_nbr,
+                                               "percentage_undetermined": percentage_undetermined,
+                                               "threshold": self.warning(),
+                                               "phix_on_lane": mean_phix_per_lane[lane_nbr]})
```
b97pla: could be nice to refactor out this code duplicaton

b97pla (approved):
looks good afaict, nice work! :+1:

johandahlberg:
Fixed your comment. Feel free to merge.

b97pla:
:+1:

PR #72, opened 2018-09-25
Reads per SampleName rather than SampleId

matrulda:
This PR modifies the `reads_per_sample_handler` to count reads per SampleName rather than SampleId. 

A new Stats.json file, that contains samples with multiple sampleIds, has been added to the test resources.


checkQC/handlers/reads_per_sample_handler.py
```diff
@@ -32,12 +32,19 @@ def check_qc(self):
         for lane_dict in self.conversion_results:
             lane_nbr = int(lane_dict["LaneNumber"])
             lane_demux = lane_dict["DemuxResults"]
-            nbr_of_samples = len(lane_demux)
+            total_reads = {}
```
johandahlberg: I would suggest using `defaultdict` here to simplify the code a bit. I.e.
```
from collections import defaultdict
total_reads = defaultdict(0)
```
Then you can omit the if statement and only do `total_reads[sample_name] += sample_id_info["NumberReads"] / pow(10, 6)`.

checkQC/handlers/reads_per_sample_handler.py
```diff
@@ -32,12 +32,19 @@ def check_qc(self):
...
 
             for sample_id_info in lane_demux:
+                sample_name = sample_id_info["SampleName"]
+                if sample_name in total_reads:
+                    total_reads[sample_name] += sample_id_info["NumberReads"] / pow(10, 6)
+                else:
+                    total_reads[sample_name] = sample_id_info["NumberReads"] / pow(10, 6)
+
+            nbr_of_samples = len(total_reads.keys())
+            for sample in total_reads:
```
johandahlberg: I'd suggest using `for sample, sample_total_reads in total_reads.items(): ...` here to get the key and the value at the same time.

johandahlberg (commented):
Looks good. I have some minor comments that you can have a look at if you like.

checkQC/handlers/reads_per_sample_handler.py
matrulda: Cool, will do!

checkQC/handlers/reads_per_sample_handler.py
matrulda: Ah ofc, thanks!

matrulda:
Have another look now @johandahlberg . Should I update version number too?

johandahlberg:
Yes. The version number should be updated. I'm kind of torn between what type of update to make though. Either we should view it as a bug fix release (i.e. this is the behavior we were always intending), or we should do a major release for this, because it is changing the existing behavior/API. If we go with the later I think we should hold of on merging this for now, and merge this and #70 together and make it a 2.0 release. What do you think?

matrulda:
I'm also unsure. I'm leaning towards major release, but this feature is really needed in our production environment. Could we merge this and make another pre-release?

b97pla:
FWIW, I'd vote for major release. Especially for external people using the software, this might have implications that should be highlighted.

johandahlberg:
Ok. Great. Then lets update to the version to `v2.0.0-rc1` amd go do a pre-release that you can validate if you think that is necessary and then release version `v2.0.0` once that is done.

matrulda:
Feel free to merge @johandahlberg @b97pla

PR #77, opened 2018-11-15
iSeq instrument names start with FS

johandahlberg:
The documentation regarding iSeq naming conventions was incorrect, the naming convention for iSeqs should be that they start with FS.


matrulda:
:+1:

PR #79, opened 2018-11-27
NovaSeq SP support

monikaBrandt:
In this PR quality thresholds for NovaSeq SP are added to the configuration file. Since no testdata is available I did the assumption that RunParameters.xml will look the same for all NovaSeq runs. However, this is not confirmed by Illumina Tech support. Please, do not merge before this is confirmed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ Instrument types supported in checkQC are the following:
 
 Install instructions
 --------------------
-CheckQC **requires Python 3.5** (or higher to run). CheckQC can be installed with pip.
+CheckQC **requires Python 3.6** (or higher to run). CheckQC can be installed with pip.
 
 ```
 pip install checkqc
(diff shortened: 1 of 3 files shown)


johandahlberg (approved):

johandahlberg:
Looks good to me. I haven't tried in on a SP flowcell, but it guess you have? I'm good to merge unless you want @MatildaAslin to also have a look. Great work!

monikaBrandt:
According to Illumina Tech support they can't tell us anything about the NovaSeq SP run output with certainty. However, they answer that it should be safe to make the assumption that the output is the same as for other NovaSeq runs.

monikaBrandt:
@johandahlberg There is no test data available for NovaSeq SP. We have to decide if we want to merge anyway. See previous comments.

johandahlberg:
Well I think that we can merge this now, and then make adjustments as needed later. Do you want me to wait for @MatildaAslins review, or should I merge?

matrulda (approved):
Nice! Checked the values against INS-00123, looks correct. :+1:

PR #81, opened 2019-03-07
NovaSeq SP 101bp

sarek928:
Added values for NovaSeq SP read length 101bp.


johandahlberg:
Don't forget to increase the version if you want to make a new release with this, which I assume you want. :smile:

sarek928:
Fixat!

matrulda (changes requested):
Noice! Looks great, but I think the error rate threshold should be 2.

PR #83, opened 2019-06-12
CheckQC will ignore white-listed indexes

johandahlberg:
This makes CheckQC ignore any index that is listed in `white_listed_indexes`, in the configuration. These can be specified as regular expressions, so it should be fairly flexible how you to specify them.

Which indexes would you like to have as default in the whitelist, @MatildaAslin and @sarek928?

### DIFF
--- a/checkQC/default_config/config.yaml
+++ b/checkQC/default_config/config.yaml
@@ -22,7 +22,7 @@ default_handlers:
     # N{3,} will match three or more Ns.
     white_listed_indexes:
       - .*N.*
-      - G{8,}
+      - G{6,}
 
 hiseq2500_rapidhighoutput_v4:
   51-71:
--- a/checkQC/default_config/config.yaml
+++ b/checkQC/default_config/config.yaml
@@ -15,7 +15,7 @@ default_handlers:
     error: 9 # <% Phix on lane> + < value as %>
   - name: UnidentifiedIndexHandler
     significance_threshold: 1 # % of reads in unidentified
-    # Indexes which are will not cause an error even if they occur
+    # Indexes which are white-listed will only cause a warning even if they occur
     # above the significance level.
     # They will be matched like regular expressions,
     # so e.g. NNN will match exactly three NNNs, while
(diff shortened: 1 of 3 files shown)


checkQC/default_config/config.yaml
```diff
@@ -12,11 +10,19 @@ parser_configurations:
...
+  - name: UnidentifiedIndexHandler
+    significance_threshold: 1 # % of reads in unidentified
+    # Indexes which are will not cause an error even if they occur
+    # above the significance level.
+    # They will be matched like regular expressions,
+    # so e.g. NNN will match exactly three NNNs, while
+    # N{3,} will match three or more Ns.
+    white_listed_indexes:
+      - .*N.*
+      - G{8,}
```
matrulda: I would suggest `G{6,}` as default instead of 8.

matrulda (approved):
LGTM! 
One thought: It would be nice if the white-listed tags weren't totally ignored, but instead listed as a warning including the count. For example, in cases where the Undetermined Percentage Handler throws an error due to 50 % Undetermined, it would be helpful to see that the indexes that stands out have `N` in them. This does however change the behaviour of the handler a bit, so maybe it's better to include in another PR.

checkQC/default_config/config.yaml
```diff
@@ -12,11 +10,19 @@ parser_configurations:
...
-      warning: unknown
-      error: 9 # <% Phix on lane> + < value as %>
-    - name: UnidentifiedIndexHandler
-      significance_threshold: 1 # % of reads in unidentified
+  - name: UndeterminedPercentageHandler
+    warning: unknown
+    error: 9 # <% Phix on lane> + < value as %>
+  - name: UnidentifiedIndexHandler
+    significance_threshold: 1 # % of reads in unidentified
+    # Indexes which are will not cause an error even if they occur
```
matrulda: Sorry for noticing this now, but I guess a word is missing? "Indexes which are whitelisted"?

johandahlberg: np. Will fix that.

matrulda:
Great work!

PR #84, opened 2019-10-08
DEVELOP-641 Downgrade error level

matrulda:
This PR adds to option to downgrade errors to warnings for specific handlers. 

TODO: Make it possible to use this parameter through the web app. 

Sharing this early so you can let me know if you have any thoughts on the solution.

**EDIT: 191009**
Downgrade error logic was moved to config.py and the functionality has been added to the web app too. This is ready for merge.

### DIFF
--- a/checkQC/app.py
+++ b/checkQC/app.py
@@ -24,9 +24,10 @@ log = logging.getLogger(__name__)
 @click.command("checkqc")
 @click.option("--config", help="Path to the checkQC configuration file", type=click.Path())
 @click.option('--json', is_flag=True, default=False, help="Print the results of the run as json to stdout")
+@click.option('--downgrade-error', type=str, multiple=True, help="Downgrade errors to warnings for a specific handler, can be used multiple times")
 @click.version_option(checkqc_version)
 @click.argument('runfolder', type=click.Path())
-def start(config, json, runfolder):
+def start(config, json, downgrade_error, runfolder):
     """
     checkQC is a command line utility designed to quickly gather and assess quality control metrics from an
     Illumina sequencing run. It is highly customizable and which quality controls modules should be run
@@ -35,7 +36,7 @@ def start(config, json, runfolder):
     # -----------------------------------
     # This is the application entry point
     # -----------------------------------
-    app = App(runfolder, config, json)
+    app = App(runfolder, config, json, downgrade_error)
     app.run()
     sys.exit(app.exit_status)
 
@@ -45,10 +46,11 @@ class App(object):
     This is the main application object for CheckQC.
     """
 
-    def __init__(self, runfolder, config_file=None, json_mode=False):
+    def __init__(self, runfolder, config_file=None, json_mode=False, downgrade_error=()):
         self._runfolder = runfolder
         self._config_file = config_file
         self._json_mode = json_mode
+        self._downgrade_error = downgrade_error
         self.exit_status = 0
 
     def configure_and_run(self):
@@ -77,7 +79,8 @@ class App(object):
 
         qc_engine = QCEngine(runfolder=self._runfolder,
                              parser_configurations=parser_configurations,
-                             handler_config=handler_config)
+                             handler_config=handler_config,
+                             downgrade_error=self._downgrade_error)
         reports = qc_engine.run()
         reports["run_summary"] = run_type_summary
         self.exit_status = qc_engine.exit_status
--- a/checkQC/app.py
+++ b/checkQC/app.py
@@ -24,10 +24,10 @@ log = logging.getLogger(__name__)
 @click.command("checkqc")
 @click.option("--config", help="Path to the checkQC configuration file", type=click.Path())
 @click.option('--json', is_flag=True, default=False, help="Print the results of the run as json to stdout")
-@click.option('--downgrade-error', type=str, multiple=True, help="Downgrade errors to warnings for a specific handler, can be used multiple times")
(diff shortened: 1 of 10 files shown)


checkQC/app.py
```diff
@@ -45,10 +46,11 @@ class App(object):
     This is the main application object for CheckQC.
     """
 
-    def __init__(self, runfolder, config_file=None, json_mode=False):
+    def __init__(self, runfolder, config_file=None, json_mode=False, downgrade_error=()):
```
johandahlberg: I think it would be better to call this `downgrade_errors_for` or possibly `downgrade_errors_list`, so that it is easier to understand that this is something that happens for specific handlers, and not for all of them.

johandahlberg: I also think that it is clearer to set the default to `None`.

checkQC/handlers/qc_handler.py
```diff
@@ -248,3 +252,14 @@ def report(self):
 
         return sorted_errors_and_warnings
 
+    def downgrade_errors(self, errors_and_warnings):
+        """
+        Takes a list of error and warnings and downgrade all errors to warnings.
```
johandahlberg: Perhaps it would be clearer to have the downgrading as an option on the handler, and then pass it in when it is instantiated? This way you wouldn't need to do any extra parsing under `_compile reports`, but you could handle all the downgrading in the `QCHandler`. What do you think?

johandahlberg: I guess you could then pass it along in the class configuration?

checkQC/qc_engine.py
```diff
@@ -100,9 +102,14 @@ def _run_parsers(self):
     def _compile_reports(self):
         reports = {"exit_status": 0}
         for handler in self._handlers:
-            handler_report = handler.report()
+            handler_name = type(handler).__name__
```
johandahlberg: This is the code I'm thinking that we can get rid of if we follow the approach I suggest above. A neat thing about that solution is that it preserves the separation of concerns between the layers here. I.e. the handler alone is responsible for figuring out if something should be an error or a warning, and the engine can just parse it as before.

tests/test_app.py
```diff
@@ -19,6 +19,11 @@ def test_run_json_mode(self):
         # The test data contains fatal qc errors
         self.assertEqual(app.run(), 1)
 
+    def test_run_downgrade_error(self):
+        app = App(runfolder=self.RUNFOLDER, downgrade_error="ReadsPerSampleHandler")
+        # Test data should not produce fatal qc errors anymore
+        self.assertEqual(app.run(),0)
```
johandahlberg: Nit picking, but there is a missing white-space here in, i.e. it should be: `self.assertEqual(app.run(), 0)`

johandahlberg (changes requested):
Looking good! I had one idea for how to separate the layers a bit more, however that should be relatively easy, with the work that you've already done. Awesome :pancakes:!

checkQC/app.py
matrulda: I updated the variable name, good suggestion! 
I did not update the default value, an iterable is needed. I thought it would be cleaner to do this than to convert the None value to a empty tuple in the code. Maybe you can comment again on the new commits if you disagree.

tests/test_app.py
matrulda: Nit picking is good, fixed!

checkQC/app.py
johandahlberg: No, this looks good. I wanted to avoid the problem described here: https://effbot.org/zone/default-values.htm However, I had forgotten that tuple in python is immutable, so in this case it should be ok, unless I misunderstand something. So let's stick with this. :+1:

johandahlberg:
This looks good! I guess the next step is to bring this into the webserver as well?

matrulda:
Thanks! Yeah, working on it right now. :+1:

checkQC/web_app.py
```diff
@@ -49,6 +51,28 @@ def get(self, runfolder):
...
                                                       "analyze is available in the qc config?")
 
+    def prepare(self):
+        if self.request.body:
+            try:
+                self.args = json.loads(self.request.body)
+            except JSONDecodeError:
+                self._write_error(status_code=500, reason="Unexpected body, expected json.")
+
+    def post(self, runfolder):
```
johandahlberg: I think that we should add this functionality to the existing `GET` method. The reason for this is that we are not actually changing any state (which is what `POST` implies - more info on http verbs here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods). This means that we need to pick up the list of headers to ignore from the request parameters.

johandahlberg (changes requested):

checkQC/web_app.py
matrulda: Aha, I thought it was custom to use POST whenever you're sending data in the request, but the change of state distinction makes more sense to me. Will fix!

matrulda:
Something like this @johandahlberg ? This implementation only supports downgrading errors for one handler though, maybe that's fine? 

**EDIT**
It actually works for multiple handlers too: 
`curl http://localhost:9999/qc/<runfolder>/downgrade_errors\=\ReadsPerSampleHandler,UndeterminedPercentageHandler`

Adding a test for that.

matrulda:
@johandahlberg I tried naming of capture groups like we talked about, but it didn't work and I found little help docs (and on the internetz). The current solution works and is pretty clean, imo.

johandahlberg (approved):
Excellent! Great work!

PR #87, opened 2019-12-13
Option to use closest read length

matrulda:
This PR adds an option to use the closest read length found in the config, if the one used isn't found. In case of a tie between two read length, the longer one (with stricter QC criteria) will be used.


matrulda:
Oops, forgot to commit a new test config. Coming up!

PR #92, opened 2020-10-13
Update QC thresholds and remove unnecessary comments

matrulda:
Updated QC thresholds according to changes in INS-00123. Also, only kept threshold comments for one read length to reduce duplication. 

Tested on a Miseq with read length 301 bp and I could confirm that the new thresholds was used.


sarek928 (approved):
The thresholds match those in INS-00123. Nice touch with the removal of the comments, I don't think they were necessary. Great work

PR #93, opened 2020-10-13
Remove unused requirements file

matrulda:
When trying to verify if the changes in #91 had fixed the API docs I found another build problem. It was caused by the `requirements/readthedocs` not being up to date. I have now changed the settings on readthedocs so that `requirements/prod` is used instead, hence we do not need this file anymore. 

Building of docs are now successful and it contains the API section 🎉


johandahlberg:
:confetti_ball:

PR #101, opened 2021-11-25
Use PyYAML's safe_load instead of load

slohse:
PyYAML version 6.0 changed the `load()` function to [require a second argument](https://github.com/yaml/pyyaml/issues/576), a Loader. There are several different loaders, some of them having serious security implications. So to force devs to consciously make a choice of loader, they changed the signature of the `load` function.
Alternatively the `safe_load` function can be used, which is what I did.


matrulda (approved):
Nice! :+1:

PR #102, opened 2022-07-05
DEVELOP-2147: Add support for Miseq Nano and Micro

matrulda:
This PR adds support for special MiSeq flowcells (nano and micro) that generates less data than the standard one.

@mikaelamo It would be nice if you could review the changes in config.yaml since you are familiar with INS-00123. :)

### DIFF
--- a/checkQC/run_type_recognizer.py
+++ b/checkQC/run_type_recognizer.py
@@ -101,16 +101,36 @@ class MiSeq(IlluminaInstrument):
     @staticmethod
     def reagent_version(runtype_recognizer):
         """
-        Find the reagent version used for this run, as MiSeqs can have multiple
-        different reagent kit versions.
+        Find the reagent kit version (and flowcell mode if applicable) for this run
 
-        :returns: reagent version of format v[number] e.g. v3
+        :returns: reagent version in format v[reagent kit version] or [flowcell mode]_v[reagent kit version],
+         e.g. v3 or nano_v2
         """
-        try:
-            reagent_version = runtype_recognizer.run_parameters["RunParameters"]["ReagentKitVersion"]
-            return reagent_version.replace("Version", "v")
-        except KeyError:
-            raise ReagentVersionUnknown("No reagent version specified for this instrument type")
+        def _reagent_kit_version(runtype_recognizer):
+            try:
+                reagent_version = runtype_recognizer.run_parameters["RunParameters"]["ReagentKitVersion"]
+                return reagent_version.replace("Version", "v")
+            except KeyError:
+                raise ReagentVersionUnknown("No reagent version specified for this instrument type")
+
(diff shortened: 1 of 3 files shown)


checkQC/run_type_recognizer.py
```diff
@@ -101,16 +101,36 @@ def name():
...
+
+        def _flowcell_type(runtype_recognizer):
+            try:
+                tiles_per_swath = int(runtype_recognizer.run_parameters["RunParameters"]["Setup"]["NumTilesPerSwath"])
+                if tiles_per_swath == 2:
+                    return "nano"
+                elif tiles_per_swath == 4:
+                    return "micro"
+                elif tiles_per_swath >= 14:
+                    return "standard"
```
Aratz, line 124: Is it possible that `tiles_per_swath` does not match any of these values? In this case maybe you could raise the `ReagentVersionUnknown` exception?

Aratz (approved):
Looks good to me :+1: I just left one question in the code.

mikaelamo (approved):
Looks correct in regards to INS-00123.

checkQC/run_type_recognizer.py
matrulda: Ah good catch, I think it shouldn't be possible with the current MiSeq configurations, but Illumina might add something new in the future. Will fix!

matrulda:
Thanks @mikaelamo and @Aratz for the reviews :pray: ! @Aratz if you could just check the last commit to make sure I didn't introduce anything weird, I think we are good to go.

Aratz:
Looks good, great job :+1:

Aratz:
Also I just noticed Python is now introducing [match statements](https://docs.python.org/3/tutorial/controlflow.html?highlight=match%20case#match-statements) in version 3.10, so I guess in the future this will become the canonical way of writing this kind of code.

mikaelamo (approved):

matrulda:
Good input @Aratz, I'm not sure how to handle the `>= 14` condition using match case though. :thinking: Do you have an idea? Otherwise I think I'll keep it like this.

Aratz:
I'd write `case x if x >= 14:`, but I think it's good as it is now, we don't use Python 3.10 yet do we?

matrulda:
Ah ofc, thanks, @Aratz ! I went ahead and upgraded to py3.10 since it required very few changes.

matrulda:
Hmm, I had to update the ubuntu version for travis to be able to download py3.10. We should make sure that we can install py3.10 on the biotanks before including those changes in the next release.

Aratz (approved):
Looks good to me :+1:

matrulda:
I removed the py3.10 commits, our old servers can't handle py3.10. Will add those changes in a separate PR so that we can release these changes first.

PR #106, opened 2023-05-09
Develop 2548 add nova seq x support

Aratz:
This PR updates CheckQC's dependencies, and specifically `interop`, which is needed to process the new runfolder.

It also makes instrument detection case insensitive and adds tests for that.

### DIFF
--- a/checkQC/handlers/unidentified_index_handler.py
+++ b/checkQC/handlers/unidentified_index_handler.py
@@ -10,24 +10,28 @@ from checkQC.exceptions import ConfigurationError
 
 class UnidentifiedIndexHandler(QCHandler):
     """
-    The UnidentifiedIndexHandler will try to identify if an index is represented at to high a level in unidentified
-    reads, and if that is the case try to pinpoint why that is.
+    The UnidentifiedIndexHandler will try to identify if an index is
+    represented at to high a level in unidentified reads, and if that is the
+    case try to pinpoint why that is.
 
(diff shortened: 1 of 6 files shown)


matrulda:
Looks good! I do however get this error when trying to build the docs: 
```
$ sphinx-autobuild . _build/html
Running Sphinx v7.0.0

Configuration error:
There is a programmable error in your configuration file:

Traceback (most recent call last):
  File "/home/matilda/.pyenv/versions/checkqc/lib/python3.8/site-packages/sphinx/config.py", line 354, in eval_config_file
    exec(code, namespace)  # NoQA: S102
  File "/home/matilda/workspace/checkQC/docs/conf.py", line 25, in <module>
    from recommonmark.parser import CommonMarkParser
ModuleNotFoundError: No module named 'recommonmark'

Command exited with exit code: 2
```

It seems like looks like it solved by installing `recommonmark` with pip. So perhaps we need to add that to the requirements, weird that it's not included with Sphinx thoug :thinking: Maybe you can see if you can reproduce this so it's not only a thing in my environment. 

I've tried it in a virtual env with py3.8.13.

Aratz:
Looks like `recommonmark` is not maintained anymore (https://github.com/readthedocs/recommonmark) and Sphynx recommends `myst-parser` instead (https://www.sphinx-doc.org/en/master/usage/markdown.html) so I'm gonna adapt the configuration for that.

matrulda (approved):
Great! :+1:

PR #108, opened 2023-10-19
Update to py3.10

matrulda:
Changes: 
- Added features supported from python 3.10 and also enforce usage of that version in the setup.py. Unfortunately we can't allow newer versions since https://github.com/Illumina/interop only has python 3.10 support yet.
- Removed travis file and added GHA workflow to run tests and upload code coverage file instead. I excluded automatic upload to PyPI, will leave that for another PR. 
- Updated the docs with regards to the new python requirement. Also removed part about PyPI upload since it is not relevant anymore and we have docs for that in-house.
  - I would like to go through the documentation entirely since I think some things might need an update and do something about the duplicate information in README and docs. Also reconsider how much we should include in the docs. But that's a separate project.
- Bumped myst-parser requirement to be able to build the docs.


Aratz (approved):
LGTM ✅

PR #111, opened 2023-12-04
Update NovaSeqX reagent recognizer

Aratz:
The format of `RunParameters.xml` changed slightly in the last instrument update for NovaSeqX.

This PR updates CheckQC to also be able to handle this new format.


matrulda:
Looks good to me! :+1:

PR #112, opened 2024-03-14
DATAOPS-635: check index q30

nkongenelly:
UpdateD Q30 handler so that index Q30 also is checked

**Verification Protocol:**
[https://snpseq.atlassian.net/wiki/spaces/AR/pages/3413377025/WIP+Verification+of+checkqc+v4.0.1]

### DIFF
--- a/checkQC/handlers/q30_handler.py
+++ b/checkQC/handlers/q30_handler.py
@@ -32,17 +32,21 @@ class Q30Handler(QCHandler):
             lane_nbr = int(error_dict["lane"])
             read = error_dict["read"]
             percent_q30 = error_dict["percent_q30"]
-
+            is_index_read = error_dict["percent_q30"]
+            read_or_index_text = "index read" if is_index_read else "read"
+            
             if self.error() != self.UNKNOWN and percent_q30 < self.error():
-                yield QCErrorFatal("%Q30 {:.2f} was too low on lane: {} for read: {}".format(percent_q30,
+                yield QCErrorFatal("%Q30 {:.2f} was too low on lane: {} for {}: {}".format(percent_q30,
                                                                                              lane_nbr,
+                                                                                             read_or_index_text,
                                                                                              read),
                                    ordering=lane_nbr,
                                    data={"lane": lane_nbr, "read": read,
                                          "percent_q30": percent_q30, "threshold": self.error()})
             elif self.warning() != self.UNKNOWN and percent_q30 < self.warning():
-                yield QCErrorWarning("%Q30 {:.2f} was too low on lane: {} for read: {}".format(percent_q30,
+                yield QCErrorWarning("%Q30 {:.2f} was too low on lane: {} for {}: {}".format(percent_q30,
                                                                                                lane_nbr,
+                                                                                               read_or_index_text,
                                                                                                read),
                                      ordering=lane_nbr,
                                      data={"lane": lane_nbr, "read": read,
--- a/checkQC/handlers/q30_handler.py
+++ b/checkQC/handlers/q30_handler.py
@@ -36,21 +36,37 @@ class Q30Handler(QCHandler):
             read_or_index_text = "index read" if is_index_read else "read"
             
             if self.error() != self.UNKNOWN and percent_q30 < self.error():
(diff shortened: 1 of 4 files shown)


checkQC/parsers/interop_parser.py
```diff
@@ -53,7 +53,8 @@ def run(self):
         py_interop_summary.summarize_run_metrics(run_metrics, summary)
 
         lanes = summary.lane_count()
-        reads = self.get_non_index_reads(summary)
+        reads = [read_nbr for read_nbr in range(summary.size())]
```
Aratz: Since you're getting all the reads now you can probably remove this line and replace line 62 by
```python
for read_nbr in range(summary.size()):
```

checkQC/handlers/q30_handler.py
```diff
@@ -32,17 +32,21 @@ def check_qc(self):
...
+            
             if self.error() != self.UNKNOWN and percent_q30 < self.error():
-                yield QCErrorFatal("%Q30 {:.2f} was too low on lane: {} for read: {}".format(percent_q30,
+                yield QCErrorFatal("%Q30 {:.2f} was too low on lane: {} for {}: {}".format(percent_q30,
                                                                                              lane_nbr,
+                                                                                             read_or_index_text,
                                                                                              read),
                                    ordering=lane_nbr,
                                    data={"lane": lane_nbr, "read": read,
                                          "percent_q30": percent_q30, "threshold": self.error()})
```
Aratz:
```suggestion
                yield QCErrorFatal(
                    "%Q30 {:.2f} was too low on lane: {} for {}: {}".format(
                        percent_q30,
                        lane_nbr,
                        read_or_index_text,
                        read
                    ),
                    ordering=lane_nbr,
                    data={
                        "lane": lane_nbr,
                        "read": read,
                        "percent_q30": percent_q30,
                        "threshold": self.error()
                    },
                )
```

checkQC/handlers/q30_handler.py
```diff
@@ -32,17 +32,21 @@ def check_qc(self):
...
             elif self.warning() != self.UNKNOWN and percent_q30 < self.warning():
-                yield QCErrorWarning("%Q30 {:.2f} was too low on lane: {} for read: {}".format(percent_q30,
+                yield QCErrorWarning("%Q30 {:.2f} was too low on lane: {} for {}: {}".format(percent_q30,
                                                                                                lane_nbr,
+                                                                                               read_or_index_text,
```
Aratz: Github does not allow me to make a suggestion that spans all the lines here but please apply the same formatting than above

Aratz (commented):
Nice work! Just had a couple comments

checkQC/parsers/interop_parser.py
nkongenelly: I have a question on this, kindly correct me as I am still learning. 
I see that on line 62, both the read value (i.e original_read_nbr) and index (i.e new_read_nbr) are been used.
And I think using this; `for read_nbr in range(summary.size()):` will mean we only have the index of the read. Or is the index read same as the value of each read?

checkQC/handlers/q30_handler.py
nkongenelly: I have now done this. Thanks for the reminder that if i can leave the file i am working on better, I should probably do that.

checkQC/parsers/interop_parser.py
Aratz: Good question, what I meant is that if `reads` is set to `[read_nbr for read_nbr in range(summary.size())]`, it will be something like `[0, 1, 2, 3, ..., summary.size()]`. Now when `enumerate` is applied to this list on line 62 it will produce something like `[(0, 0), (1, 1), (2, 2), ...]`. So basically `new_read_nbr` and `original_read_nbr` will always be equal.

I'm not super familiar with this code either, but my guess is that before, it was necessary to have both new and original read numbers because the index reads were discarded.

nkongenelly: Oh, yes I now understand that. Thanks for the explanation. I have now made that change.

Aratz (approved):

PR #113, opened 2024-04-04
Added index positions

nkongenelly:
To not confuse the use,  the reads now are being referred to as (in the case of paired-end reads with dual index): 
read 1, index 1, index 2, read 2
and not 
read 1, read 2, read 3, read 4

**Verification protocol**:

https://snpseq.atlassian.net/wiki/spaces/AR/pages/3413377025/Verification+of+checkqc+v4.0.1


checkQC/handlers/q30_handler.py
```diff
@@ -32,37 +32,39 @@ def check_qc(self):
...
                         "read": read,
+                        "index": index,
                         "percent_q30": percent_q30, 
                         "threshold": self.error()
                     }
                 )
             elif self.warning() != self.UNKNOWN and percent_q30 < self.warning():
                 yield QCErrorWarning(
-                    "%Q30 {:.2f} was too low on lane: {} for {}: {}".format(
+                    "%Q30 {:.2f} was too low on lane: {} for read: {}, index: {}".format(
```
matrulda: Ah sorry if I was unclear, what I meant was: 
For the first non-index read the message should be: 
`%Q30 {:.2f} was too low on lane: {} for read: 1`

For the first index read it should be: 
`%Q30 {:.2f} was too low on lane: {} for index read: 1`

For the second index read (if it exists) it should be: 
`%Q30 {:.2f} was too low on lane: {} for index read: 2`

For the second non-index read (if it exists) it should be: 
`%Q30 {:.2f} was too low on lane: {} for read: 2`

checkQC/parsers/interop_parser.py
```diff
@@ -68,7 +68,7 @@ def run(self):
                 self._send_to_subscribers(("error_rate",
                                         {"lane": lane+1, "read": read_nbr+1, "error_rate": error_rate}))
                 self._send_to_subscribers(("percent_q30",
-                                        {"lane": lane+1, "read": read_nbr+1, "percent_q30": percent_q30, "is_index_read":is_index_read}))
+                                        {"lane": lane+1, "read": read_nbr+1, "percent_q30": percent_q30, "index":read_nbr+1 if is_index_read else "-"}))
```
matrulda: Given my other comment this should be changed back to `"is_index_read": is_index_read`

checkQC/handlers/q30_handler.py
matrulda: So you previous implementation was kind of correct but the index (that is, the "number ID" not the index read, this word is so confusing in this context :facepalm: ) of the reads need to be corrected. 

It makes no sense to talk about index read 3, when there are only 2 index reads. 

Let me know if this is unclear.

matrulda (changes requested):
I'm sorry, I think my description was unclear. See comments and let me know if you have any questions.

nkongenelly:
Hi @matrulda , kindly confirm (maybe side to side) if the current version in master is what is expected. I think what you have explained now is already in master if I am not wrong

checkQC/handlers/q30_handler.py
nkongenelly: Thank you for the clarification, I have a question on where to use the "number ID"...should it be like this now?

For the first non-index read the message should be:
`%Q30 {:.2f} was too low on lane: {} for read: 1`

For the first index read it should be:
`%Q30 {:.2f} was too low on lane: {} for number ID: 1`

For the second index read (if it exists) it should be:
`%Q30 {:.2f} was too low on lane: {} for number ID: 2`

For the second non-index read (if it exists) it should be:
`%Q30 {:.2f} was too low on lane: {} for read: 2`

matrulda:
Will do!

nkongenelly:
> Will do!

Actually sorry for that. I had not seen you had mentioned that you checked the older version and suggested the changes from there. 
I have now commented on your previous comment just to make sure I understand how to use the ""number ID"

checkQC/handlers/q30_handler.py
matrulda: Ah no, I didn't mean it literally, just tried to say index without mixing it up index reads. This is still accurate: https://github.com/Molmed/checkQC/pull/113#discussion_r1551696310

matrulda:
> > Will do!
> 
> Actually sorry for that. I had not seen you had mentioned that you checked the older version and suggested the changes from there. I have now commented on your previous comment just to make sure I understand how to use the ""number ID"

No worries! Just to be sure I also checked the changes in the last PR (merged to master) again and I'm pretty sure my suggestion is not implemented. There is a test that verifies that: 
`self.assertEqual(f"{errors_and_warnings}", "[Fatal QC error: %Q30 50.00 was too low on lane: 1 for index read: 3]")` 

But since it is the first and only index, this is what we want to assert: 
`self.assertEqual(f"{errors_and_warnings}", "[Fatal QC error: %Q30 50.00 was too low on lane: 1 for index read: 1]")`

nkongenelly:
@matrulda Thanks for explaining to me until I understood how the indexed and non-indexed reads should show. I have now corrected that and ready for your review when you get time.

checkQC/handlers/q30_handler.py
```diff
@@ -28,12 +28,18 @@ def collect(self, signal):
...
+        non_index_count = 0
         for error_dict in self.error_results:
             lane_nbr = int(error_dict["lane"])
-            read = error_dict["read"]
             percent_q30 = error_dict["percent_q30"]
-            is_index_read = error_dict["percent_q30"]
+            is_index_read = error_dict.get("is_index_read", False)
             read_or_index_text = "index read" if is_index_read else "read"
+            # Differentiate read values for indexed from non-indexed reads
+            index_count = index_count + 1 if is_index_read else index_count
```
matrulda: Suggestion to make it a little bit shorter: 
`index_count += 1 if is_index_read else 0`

matrulda (approved):
Looks good! Just added a minor suggestion.

checkQC/handlers/q30_handler.py
nkongenelly: Thanks for that review. I have now updated the code.

Maybe you can have a final look just incase then when ready I can merge

checkQC/handlers/q30_handler.py
```diff
@@ -28,12 +28,18 @@ def collect(self, signal):
...
+        non_index_count = 0
         for error_dict in self.error_results:
             lane_nbr = int(error_dict["lane"])
-            read = error_dict["read"]
             percent_q30 = error_dict["percent_q30"]
-            is_index_read = error_dict["percent_q30"]
+            is_index_read = error_dict.get("is_index_read", False)
             read_or_index_text = "index read" if is_index_read else "read"
+            # Differentiate read values for indexed from non-indexed reads
+            index_count += 1 if is_index_read else index_count
```
matrulda: I think you need to change the last `index_count` to 0. We want the counter to NOT be increased if we loop over a non index read, not to add the current `index_count`. I think this would have yield an inaccurate count if the `TestQ30Handler.setUp` looked like this: 

```
        value_1 = {"lane": 1, "read": 1, "percent_q30": 82}
        value_2 = {"lane": 1, "read": 2, "percent_q30": 50, "is_index_read": True}
        value_3 = {"lane": 1, "read": 3, "percent_q30": 82}
        value_4 = {"lane": 1, "read": 4, "percent_q30": 50, "is_index_read": True}
```

matrulda (changes requested):

checkQC/handlers/q30_handler.py
nkongenelly: Thanks for this catch. I have learnt something new here. I should have tested that part before submitting it. 

I have now corrected the above code as suggested.

matrulda (approved):
Great! :+1:

PR #114, opened 2024-05-16
Fixed reads given multiple lanes

nkongenelly:
The code below is a fix for the below issue:

Wrong reads were found when multiple lanes are present

E.g:
    - data:
        lane: 7
        percent_q30: 83.2435073852539
        read: 14
        threshold: 85
      message: '%Q30 83.24 was too low on lane: 7 for read: 14'
      type: warningFYI: I've found a bug in CheckQC that needs to be solved asap. Me and

### DIFF
--- a/checkQC/handlers/q30_handler.py
+++ b/checkQC/handlers/q30_handler.py
@@ -30,17 +30,24 @@ class Q30Handler(QCHandler):
 
         index_count = 0
         non_index_count = 0
+        lane_no = 1
+
         for error_dict in self.error_results:
             lane_nbr = int(error_dict["lane"])
+            # restart counters for different lanes
+            if lane_nbr != lane_no:
+                index_count = 0
+                non_index_count = 0
+                lane_no = lane_nbr
+
             percent_q30 = error_dict["percent_q30"]
             is_index_read = error_dict.get("is_index_read", False)
             read_or_index_text = "index read" if is_index_read else "read"
             # Differentiate read values for indexed from non-indexed reads
             index_count += 1 if is_index_read else 0
             non_index_count += 1 if not is_index_read else 0
-            
             read = index_count if is_index_read else non_index_count
-            
+
             if self.error() != self.UNKNOWN and percent_q30 < self.error():
                 yield QCErrorFatal(
                     "%Q30 {:.2f} was too low on lane: {} for {}: {}".format(
--- a/checkQC/handlers/q30_handler.py
+++ b/checkQC/handlers/q30_handler.py
@@ -27,13 +27,11 @@ class Q30Handler(QCHandler):
             self.error_results.append(value)
 
     def check_qc(self):
-
-        index_count = 0
-        non_index_count = 0
-        lane_no = 1
+        lane_no = 0
 
         for error_dict in self.error_results:
             lane_nbr = int(error_dict["lane"])
+
             # restart counters for different lanes
             if lane_nbr != lane_no:
                 index_count = 0
--- a/tests/handlers/test_q30_handler.py
+++ b/tests/handlers/test_q30_handler.py
@@ -56,13 +56,24 @@ class TestQ30Handler(HandlerTestBase):
         errors_and_warnings = list(self.q30_handler.check_qc())
         self.assertEqual(len(errors_and_warnings), 8)
 
-        class_names = self.map_errors_and_warnings_to_class_names(errors_and_warnings)
-        self.assertListEqual(
-            class_names,
-            [
-                "QCErrorFatal", "QCErrorFatal", "QCErrorFatal", "QCErrorFatal", "QCErrorFatal", "QCErrorFatal", "QCErrorFatal", "QCErrorFatal",
-            ],
+        self.assertEqual(
+            str(errors_and_warnings),
+            str("[Fatal QC error: %Q30 82.00 was too low on lane: 1 for read: 1, Fatal QC error: %Q30 50.00 was too low on lane: 1 for index read: 1, Fatal QC error: %Q30 50.00 was too low on lane: 1 for index read: 2, Fatal QC error: %Q30 60.00 was too low on lane: 1 for read: 2, Fatal QC error: %Q30 90.00 was too low on lane: 2 for read: 1, Fatal QC error: %Q30 50.00 was too low on lane: 2 for index read: 1, Fatal QC error: %Q30 50.00 was too low on lane: 2 for read: 2, Fatal QC error: %Q30 40.00 was too low on lane: 2 for read: 3]")
         )
+
+    def test_max_read(self):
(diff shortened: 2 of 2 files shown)


tests/handlers/test_q30_handler.py
```diff
@@ -9,43 +9,79 @@ class TestQ30Handler(HandlerTestBase):
...
     def test_all_is_fine(self):
-        qc_config = {'name': 'Q30Handler', 'error': 70, 'warning': 80}
+        qc_config = {"name": "Q30Handler", "error": 70, "warning": 80}
         self.set_qc_config(qc_config)
         errors_and_warnings = list(self.q30_handler.check_qc())
         self.assertIn("index read", f"{errors_and_warnings}")
-        self.assertEqual(f"{errors_and_warnings}", "[Fatal QC error: %Q30 50.00 was too low on lane: 1 for index read: 1]")
+        self.assertEqual(
+            f"{errors_and_warnings}",
+            "[Fatal QC error: %Q30 50.00 was too low on lane: 1 for index read: 1, Fatal QC error: %Q30 50.00 was too low on lane: 1 for index read: 2, Fatal QC error: %Q30 60.00 was too low on lane: 1 for read: 2, Fatal QC error: %Q30 50.00 was too low on lane: 2 for index read: 1, Fatal QC error: %Q30 50.00 was too low on lane: 2 for read: 2, Fatal QC error: %Q30 40.00 was too low on lane: 2 for read: 3]",
```
matrulda: I realize I should have reacted to this during your last PR, but I guess I missed it: The purpose of this test was originally to test the case where the handler does not report any warnings or errors (hence the name all_is_fine). See         [self.assertEqual(errors_and_warnings, [])](https://github.com/Molmed/checkQC/blob/639d3e2c0555ae3b4810fdef542fe22ba48ade0c/tests/handlers/test_q30_handler.py#L27). I think the qc_config for this case should be updated so that is the case, for example: 
```
qc_config = {"name": "Q30Handler", "error": 20, "warning": 30}
```

It would also be fine to restructure the tests, I think the most important thing is that the test name reflect what we are testing.

tests/handlers/test_q30_handler.py
```diff
@@ -9,43 +9,79 @@ class TestQ30Handler(HandlerTestBase):
...
+        self.assertListEqual(
+            class_names,
+            [
+                "QCErrorFatal", "QCErrorFatal", "QCErrorFatal", "QCErrorFatal", "QCErrorFatal", "QCErrorFatal", "QCErrorFatal", "QCErrorFatal",
+            ],
+        )
 
-if __name__ == '__main__':
+    def test_error_low_qc_marks(self):
+        qc_config = {"name": "Q30Handler", "error": 40, "warning": 60}
```
matrulda: I would suggest removing this test and add this config value to `test_warning` instead.

tests/handlers/test_q30_handler.py
```diff
@@ -9,43 +9,79 @@ class TestQ30Handler(HandlerTestBase):
...
+        qc_config = {"name": "Q30Handler", "error": 70, "warning": 80}
         value_1 = {"lane": 1, "read": 1, "percent_q30": 82}
-        value_2 = {"lane": 1, "read": 2, "percent_q30": 90}
-        value_3 = {"lane": 1, "read": 3, "percent_q30": 50, "is_index_read": True}
+        value_2 = {"lane": 1, "read": 2, "percent_q30": 50, "is_index_read": True} 
+        value_3 = {"lane": 1, "read": 3, "percent_q30": 50, "is_index_read": True} 
+        value_4 = {"lane": 1, "read": 4, "percent_q30": 60}
+        value_5 = {"lane": 2, "read": 1, "percent_q30": 90}
+        value_6 = {"lane": 2, "read": 2, "percent_q30": 50, "is_index_read": True}
+        value_7 = {"lane": 2, "read": 3, "percent_q30": 50, "is_index_read": False}
```
matrulda:
```suggestion
        value_7 = {"lane": 2, "read": 3, "percent_q30": 50}
```
Since `is_index_read = False` is default there is no reason to write this out right?

checkQC/handlers/q30_handler.py
```diff
@@ -30,17 +30,24 @@ def check_qc(self):
 
         index_count = 0
         non_index_count = 0
+        lane_no = 1
```
matrulda:
```suggestion
        lane_no = 0
```
If lane is set to one here, I think line 31-32 could be removed. Then the counters will be set in the for loop during the first iteration. If I'm not missing something. Let me know if you agree.

matrulda (commented):
Great work! Left some comments for you.

checkQC/handlers/q30_handler.py
nkongenelly: Yes, I agree, that is brilliant.

nkongenelly:
@matrulda , thanks for the review. I have now made the suggested changes.

matrulda (approved):
Looks good, great work! :+1:

matrulda:
I realized that these tests would have passed with the original code. Before merging we should add another test that checks that no read reported is greater than 4. It would also be fine to, for example, update the warning test so that we check all messages instead of looking at the classes.

nkongenelly:
@matrulda Thanks for that comment, yes the suggested additional tests are great and I have now added them.

tests/handlers/test_q30_handler.py
```diff
@@ -56,14 +56,25 @@ def test_error(self):
...
 
-        class_names = self.map_errors_and_warnings_to_class_names(errors_and_warnings)
-        self.assertListEqual(
-            class_names,
-            [
-                "QCErrorFatal", "QCErrorFatal", "QCErrorFatal", "QCErrorFatal", "QCErrorFatal", "QCErrorFatal", "QCErrorFatal", "QCErrorFatal",
-            ],
+        self.assertEqual(
+            str(errors_and_warnings),
+            str("[Fatal QC error: %Q30 82.00 was too low on lane: 1 for read: 1, Fatal QC error: %Q30 50.00 was too low on lane: 1 for index read: 1, Fatal QC error: %Q30 50.00 was too low on lane: 1 for index read: 2, Fatal QC error: %Q30 60.00 was too low on lane: 1 for read: 2, Fatal QC error: %Q30 90.00 was too low on lane: 2 for read: 1, Fatal QC error: %Q30 50.00 was too low on lane: 2 for index read: 1, Fatal QC error: %Q30 50.00 was too low on lane: 2 for read: 2, Fatal QC error: %Q30 40.00 was too low on lane: 2 for read: 3]")
```
matrulda, line 61: It would be nice to re-format this so that it is easier to read, but let's keep it like this for now.

matrulda (approved):
Awesome work!

PR #115, opened 2024-06-04
Dataops 638 q30 drop per cycle

monikaBrandt:
This PR includes collecting data from Interop-files for a new feature in CheckQC. The new feature is to also check %Q30 on a per cycle level to detect unexpected quality drops within a read. The following functionality is implemented:
- Reading additional (more detailed) quality metrics from the interop-files using `imaging` from the interop library.
- Truncate number of cycles to investigate to not include cycles with expected Q30 drops.
- Summarize Q30 values for a specific read, lane and cycle over tiles and return the result in a dictionary containing values for all selected cycles within the read.  
- Unittest to verify the described functionally were implemented in collaboration with @Aratz


checkQC/parsers/interop_parser.py
```diff
@@ -32,46 +56,135 @@ def get_non_index_reads(summary):
...
+        q30_cycle_df = pd.DataFrame(q_metrics)
+        q30_lane_read = q30_cycle_df[
+                (q30_cycle_df["Lane"] == lane_nr+1) &\
+                (q30_cycle_df["Read"] == read_nr+1)]
+
+        q30_per_cycle = {}
+        if is_index_read:
+            end_cycle = int(max(q30_lane_read["Cycle Within Read"]))
+            start_cycle = end_cycle - 5
+            #Remove the first 2 bases in the index read.
```
Aratz: Is this comment up to date? I'm not sure I understand why it says the first two bases here but the like above is `end_cycle - 5`. Does this mean this only works for indices of a specific length?

Aratz: Also, usually, comments are put before the line/section they refer to (see https://peps.python.org/pep-0008/#block-comments)

checkQC/parsers/interop_parser.py
```diff
@@ -32,46 +56,135 @@ def get_non_index_reads(summary):
...
+
+        else:
+            end_cycle = math.ceil(
+                max(q30_lane_read["Cycle Within Read"])*0.9
+            )
+
+            #Remove the last 90% of all cycles since 
+            #they are expected to drop in q30
+            start_cycle = 6
```
Aratz: Are cycles 1-indexed? It would be nice to have this either in the function's documentation or as a comment

checkQC/parsers/interop_parser.py
```diff
@@ -32,46 +56,135 @@ def get_non_index_reads(summary):
...
+
+        for cycle in range(start_cycle, end_cycle+1):
+            q30_cycle_mean = float(numpy.mean(
+                q30_lane_read[
+                    (q30_lane_read["Cycle Within Read"] == cycle)
+                ]["%>= Q30"]
+            ))
+            q30_per_cycle[cycle] = q30_cycle_mean
+
+        return q30_per_cycle
```
Aratz:
```suggestion
        return {
            cycle: q30_lane_read[
                (q30_lane_read["Cycle Within Read"] == cycle)
            ]["%>= Q30"].mean()
            for cycle in range(start_cycle, end_cycle+1)
        }
```
And remove dictionary initialization on line 103

Aratz (commented):
Nice work! I have a few questions about the function :)

checkQC/parsers/interop_parser.py
monikaBrandt: This was a way to get around bad quality in the beginning of the index read in our test run. However, this does not seem to be an issue in newer runs so I decided to keep all cycles for the index read.

checkQC/parsers/interop_parser.py
monikaBrandt: Thanks for pointing this out! In this case start_cycle/cycle/end_cycle will not be used as an index to get and element from the data frame (that is 0-indexed). Instead it is used to select data based on the values in column 'Cycle' and 'Cycle Within Read' which starts at 1. It is a bit confusing but make sense in this context. I will make a comment about it.

checkQC/parsers/interop_parser.py
monikaBrandt: I remember we had a discussion about this during our pair-programming session. For me, your suggested changes are harder for me to read. However, if you have strong opinions on this I can go ahead and change it :slightly_smiling_face:

Aratz: This is the idiomatic way of constructing a dictionary in Python.

Also, more than just turning a loop into a dict-comprehension, it addresses the following points:
- it uses the dataframe's `.mean()` instead of numpy's , and solve the float issue we have
- it removes intermediate temporary variables

Aratz (approved):

PR #117, opened 2024-09-11
DATAOPS-779: Fix undetermined percentage handler

matrulda:
This PR fixes an issue where the Undetermined handler did not report any errors. This issue was introduced when we updated the interop parser to return info about index reads. PhiX for index reads are NaN and the handler did not handle this properly. 

One reason that we missed this bug, was that the interop parser was tested with a MiSeq without index reads. I have now added a new public dataset with index reads and updated the tests accordingly. 

Validation procedure: I ran this version on a runfolder where too high % determined had been reported. % Undetermined errors were displayed as expected. I don't think any further manual testing is required.


checkQC/handlers/undetermined_percentage_handler.py
```diff
@@ -33,13 +34,12 @@ def collect(self, signal):
...
         for lane, reads in self.phix_aligned.items():
-            mean = 0
-            for read, value in reads.items():
-                mean += value / len(reads)
-            lane_and_mean_percentage_phix_aliged[lane] = mean
-        return lane_and_mean_percentage_phix_aliged
+            reads_list = list(reads.values())
+            mean_phix = 0 if all(np.isnan(reads_list)) else np.nanmean(reads_list)
+            lane_and_mean_percentage_phix_aligned[lane] = mean_phix
+        return lane_and_mean_percentage_phix_aligned
```
Aratz, line 42: Neat!

tests/parsers/test_interop_parser.py
```diff
@@ -31,77 +32,94 @@ def subscribe(self):
                     self.percent_q30_values.append(interop_stat)
                 if key == "percent_q30_per_cycle":
                     self.percent_q30_per_cycle.append(interop_stat)
+                if key == "percent_phix":
+                    self.phix_aligned_values.append(interop_stat)
```
Aratz: As it is now, only one of these statements will be executed each time, but one needs to carefully look at all the conditions to see that it's the case. This could be rewritten either using a `match` statement, or (even better) by lumping all the attributes defined in `__init__` into a dictionary using the same key that is found in `interop_stat`.

Aratz (commented):
Looks good! On top of the other comment, I think it would be nice if the results of the CI were shown in pull requests. I think this can be done by changing the following line in the ci file: `on: [push, pull_request]`.

It might not work for this PR but I think it should for all the PRs that will be created after that.

checkQC/handlers/undetermined_percentage_handler.py
matrulda: Thanks! :)

matrulda:
> Looks good! On top of the other comment, I think it would be nice if the results of the CI were shown in pull requests. I think this can be done by changing the following line in the ci file: `on: [push, pull_request]`.
> 
> It might not work for this PR but I think it should for all the PRs that will be created after that.

Ah yes, I agree. Will fix!

tests/parsers/test_interop_parser.py
matrulda: Very good point, I will address this.

matrulda:
Thanks for the comments! Using a dict made it so much cleaner :ok_hand:

Aratz (approved):

PR #118, opened 2025-02-05
Implement `QCData` class

Aratz:
In order to facilitate supporting new instruments and demultiplexers, we have decided to partly refactor CheckQC. The new class will live on the side of the existing code until it implements all its functionalities.

This PR adds a new class `QCData`, which represent the data imported from a runfolder. The plan is to implement one constructor per demultiplexer, so that all QC checks can be made based on the same data structure.

As of now it is not connected to the cli or the webapp.

OBS: this PR adds json validation to the existing app, this might require a validation round.


checkQC/config.py
```diff
@@ -33,16 +35,33 @@ def _get_config_file(config_path):
...
+                    )
+                    with open(config_path) as config_file:
+                        config = yaml.safe_load(config_file)
+            jsonschema.validate(config, schema)
+            return config
         except FileNotFoundError as e:
             log.error("Could not find config file: {}".format(e))
             raise e
+        except jsonschema.exceptions.ValidationError as e:
+            log.exception("")
```
matrulda: I see the point in not logging the exception error twice, but it could be helpful to point the user to the invalid config file, suggestion:
```suggestion
            log.exception(f"Config {config_file} failed schema validation")
```

checkQC/config.py
```diff
@@ -33,16 +35,33 @@ def _get_config_file(config_path):
...
+            else:
+                with resources.as_file(
+                        base_path / "default_config/config.yaml") as config_path:
+                    log.info(
+                        "No config file specified, "
+                        f"using default config from {config_path}."
+                    )
+                    with open(config_path) as config_file:
+                        config = yaml.safe_load(config_file)
+            jsonschema.validate(config, schema)
```
matrulda, line 57: very nice addition :clap:

checkQC/parsers/illumina.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+def _read_quality_metrics(quality_metrics_path):
+    with open(quality_metrics_path, encoding="utf-8") as csvfile:
+        return list(csv.DictReader(csvfile))
+
+
+def _read_top_unknown_barcodes(top_unknown_barcodes_path):
+    with open(top_unknown_barcodes_path, encoding="utf-8") as csvfile:
+         return  list(csv.DictReader(csvfile))
```
matrulda:
```suggestion
         return list(csv.DictReader(csvfile))
```

checkQC/qc_checkers/utils.py
```diff
@@ -0,0 +1,25 @@
...
+        - CheckQC -> check_qc
+
+    This is used to ensure backward compatibility between the CheckQC v4 and v5
+
+    NB: this method is idempotent, it will not modify qc_checker strings
+    Ex: error_rate -> error_rate
+    """
+    words = [
+        mtch.group().lower().replace('_', '')
+        #for mtch in re.finditer(r"[A-Z]?([a-z0-9]+|[A-Z0-9]+)(?=$|[A-Z]|_)", s)
```
matrulda: Remove?

tests/resources/bclconvert/200624_A00834_0183_BHMTFYTINY/Reports/SampleSheet.csv
matrulda, line 1: This does not match the Samplesheet I used when re-doing the demultiplexing, I think I forgot to update this file in my previous PR. I don't think this file is used in the tests, so I think we can remove it?

tests/resources/bclconvert/200624_A00834_0183_BHMTFYTINY/Reports/RunInfo.xml
matrulda, line 1: This file is also present on runfolder level in the test data, I think this one could be removed.

tests/test_qc_data.py
```diff
@@ -0,0 +1,268 @@
...
+                "handlers": [
+                    {"name": "mock_checker", "error": 5, "warning": 2},
+                ],
+            },
+        },
+    }
+
+
+def test_report(bclconvert_runfolder, checker_configs):
+    mock_qc_error = QCErrorFatal("Mock error message")
```
matrulda: Is this used?

matrulda (commented):
Great, left some minor comments that you can have a look at.

checkQC/config.py
Aratz: Thanks :)

tests/test_qc_data.py
Aratz: Good catch! I think it's a left-over before I moved the mocks to the `bclconvert_runfolder` fixture above. I'll remove it.

tests/resources/bclconvert/200624_A00834_0183_BHMTFYTINY/Reports/RunInfo.xml
Aratz: You're right, I think these files can be removed, I'm not sure where they come from though, maybe from the demultiplex output or something? Anyways, I'll remove them :+1:

Aratz:
Thanks for the review :pray: I've addressed your comments now :+1:

matrulda (approved):
Great, thank you!

matrulda:
Also, I agree that these changes requires a verification to make sure the config is read OK.

PR #119, opened 2025-02-06
Add NextSeq 500/550 support

lbeltrame:
I do not have access to a 500 machine, so feedback would be welcome. So far I have tested it with a couple of runs I have access to, and "seems" to work.

Defaults in config were based either on Illumina specs, or personal experience.

TODO: tests and example data (first let's see if the approach is sound)


checkQC/run_type_recognizer.py
```diff
@@ -206,6 +223,7 @@ def instrument_type(self):
             "A": NovaSeq,
             "FS": ISeq,
             "LH": NovaSeqXPlus,
+            "NB": NextSeq500
```
matrulda: I had a look at demo data from a NextSeq500 on Illumina Basespace and it looks like the prefix was `NS` in that case. `NB` seems to be the prefix for NextSeq550 machines though. Should this class be called NextSeq550 instead? Are the specs in the config based on that instrument?

checkQC/default_config/config.yaml
```diff
@@ -596,3 +596,80 @@ iseq_v1:
       - name: ReadsPerSampleHandler
         warning: unknown
         error: 3
+
+nextseq500_high:
```
matrulda, line 600: Did you leave out the ReadsPerSampleHandler for a specific reason? At our facility we set it as 0.75 * The expected Cluster PF. It is divided by the number of samples sequenced on the lane in the handler.

checkQC/run_type_recognizer.py
```diff
@@ -169,6 +169,23 @@ def reagent_version(runtype_recognizer):
...
 
+class NextSeq500(IlluminaInstrument):
+
+    @staticmethod
+    def name():
+        return "nextseq500"
+
+    @staticmethod
+    def reagent_version(runtype_recognizer):
+
```
matrulda, line 180: Please add docstring with information about the function.

matrulda (commented):
Thank you for contributing to CheckQC! This looks great, I left some comments for you. 

I think you can move forward with finding test data and adding unit tests.

checkQC/run_type_recognizer.py
lbeltrame: Thanks for confirming! As the machines are identical with regards to the sequencing hardware (the 550 has extra, but it's for processing arrays, so not relevant), I'll add a separate class for the 500.

checkQC/default_config/config.yaml
lbeltrame: The specific reason is called "overzealous copy and paste" ;) 
I was also trying to add support for the 2000 family before realizing it would need bclconvert support tacked in, so in the resulting "remove, rebase, etc." shuffle something got erased by mistake.
I'll add them back.

lbeltrame:
Thanks a lot for the review. I'll first address the comments, then grab some test data (I think there's a link somewhere in some of the issues) for unit testing.

checkQC/run_type_recognizer.py
```diff
@@ -223,7 +251,8 @@ def instrument_type(self):
             "A": NovaSeq,
             "FS": ISeq,
             "LH": NovaSeqXPlus,
-            "NB": NextSeq500
+            "NS": NextSeq500,
+            "NB": NextSeq550
```
matrulda:
```suggestion
            "NB": NextSeq550,
```

matrulda: Just a suggestion to make future additions easier to spot in the diff.

checkQC/default_config/config.yaml
```diff
@@ -606,6 +606,103 @@ nextseq500_high:
...
+      - name: ErrorRateHandler
+        allow_missing_error_rate: False
+        warning: 1.5
+        error: unknown
+
+nextseq550_high:
+  76:
+    handlers:
+      - name: ClusterPFHandler
+        warning: 60
```
matrulda: I notice that you have set this the same for both mid and high, wouldn't it be expected that the expected number of clusters differ a bit? 

Referring to https://www.illumina.com/systems/sequencing-platforms/nextseq/specifications.html, I would set this threshold to 100 (high) and 32.5 (mid). Not sure about NextSeq500 since it it looks like Illumina doesn't have information about that sequencer anymore.

checkQC/default_config/config.yaml
```diff
@@ -644,6 +747,9 @@ nextseq500_low:
       - name: Q30Handler
         warning: 80
         error: unknown
+      - name: ReadsPerSampleHandler
+        warning: unknown
+        error: 15
```
matrulda: This is not a deal breaker since the users can specify this value themselves, but I would prefer if this threshold was 45 (given that cluster PF is 60), to follow the same format as the other default values. If there is a good reason for setting it at 15 I'm not impossible to convince though :) 

See my comment about the ClusterPFHandler thresholds before addressing this.

matrulda (commented):
Looks good! Just some minor comments about the default config values. 

We do not have NextSeqs at my facility, so I do not have a strong opinions about these values, but it would be nice if they make sense for most users.

checkQC/default_config/config.yaml
lbeltrame: I thought it was a percentage, this is an absolute number? If so I think the values you propose are correct and would match my experience.

checkQC/default_config/config.yaml
lbeltrame: I made back of the envelope calculations, I guess they were wrong ;)

lbeltrame:
Thanks for the feedback, I'll try to implement the suggestions soon, I also downloaded bits off BaseSpace to be used in unit tests. I'll ask here if I encounter any problems.

matrulda:
That sounds great, thank you! :) No rush, I'm actually kind of swamped this week, so it will probably take me some time to review anyway.

checkQC/default_config/config.yaml
lbeltrame: FTR, the sequencing hardware is the same as the 550, the only capability the 550 has over the 500 is the ability to process arrays.

checkQC/default_config/config.yaml
lbeltrame: I should have adjusted them, but please double check I understood the logic correctly.

lbeltrame:
I added tests for the 550, as for the 500 I need to do more digging as some of the data on BaseSpace predates the version of `bcl2fastq` which outputted `Stats.json`. @matrulda Can you check whether they are set correctly? I ran them and all pass.

matrulda:
Looks good! @lbeltrame Just make sure to update the TODO. 

Thinking about it now, I guess you don't really have to add new test data. The unit tests are only using mocks (if I'm not mistaken). Since you already added files for NextSeq550, it could be nice to add a test for that runfolder in the integration tests (test_app or test_web_app). But I think it is fine if you do not add InterOp files etc. for NextSeq500.

lbeltrame:
So, should I just remove that TODO function? If not I'll see if I can dig up a dataset (hard, some "500" in BaseSpace are actually 550s).

matrulda:
Sorry for the slow replies, I continue to be a bit swamped. Thanks for your patience! 

No I think you should keep it (but update the mock machine tag etc.). The test data that you have added is currently not used. To make sure it is used, you could add an integration test, for example here: https://github.com/Molmed/checkQC/blob/master/tests/test_web_app.py But honestly, I don't think it is needed. As long as we don't think the interop files are formatted in another way for NextSeq, which I wouldn't expect. As long as we are confident that the machine type mappings and reagent versions are mocked correctly (e.g. RunParameters.xml always contain an entry called Chemistry that states "NextSeq High/Mid") + make sure to test it on real nextSeq runs (which I think you have done already?), there is no need for actual data. I should have realized this sooner. :facepalm: If you decide to not add an integration test, you can remove the `190126_NB551090_0289_AH5VVJBGX9` test resource.

Regarding finding a NextSeq 500 on baseSpace, I found a run called `500 v2: TruSeq Nano 550 (NA12878)`, where the Runparameters states that the RunID is `150107_NS500101_0202_A150707_NHL10_A` and the Chemistry is `NextSeq High`. Which matches what you have implemented. Is this one of those runs you think is a NextSeq 550 in disguise?

lbeltrame:
Yeah, I did test on our internal runs first. ;)
I'll fix the mock tests this week and for now I'll remove the resource (I don't have the bandwidth to add integration tests at the moment - sorry!). 
When this is merged, do you think you'll be able to cut a new release?

matrulda:
Great, that sounds good :+1: 

Yes, we will do some internal validation and then I'll make a new release :) Our group is working on bclconvert support as well, so I have to coordinate this a bit, but I think it makes sense to release the NextSeq support first and then the bclconvert in a separate release.

lbeltrame:
FTR, once bclconvert support is in, I'll work towards getting NextSeq 2000 support in (we have such an instrument)...

lbeltrame:
@matrulda This should be it, hopefully. FTR, the run you found is a true NextSeq 500, and not a 550 in disguise. I used the serial for the mock test. ;)

matrulda (approved):
Awesome! Thank you so much for this :tada:

PR #120, opened 2025-02-11
Refactor `QCData`

Aratz:
While working on connecting QCData to the main app, I realized the `QCData` class was implementing several functionalities that maybe shouldn't not belong together, such as processing CheckQC's config and processing arguments provided at runtime (such as `downgrade_errors`).

This PR implements the following changes:
- `QCData`'s purpose is now a dataclass which abstract the data available in a runfolder. As such it does not contain any information about CheckQC's configuration.
- `QCReporter` abstracts the process of analyzing data and producing report
  - the class' contructor takes CheckQC's config (namely static information that will not change unless the service is restarted).
  - The method `gather_reports` takes an instance of `QCData` as parameter, as well as all arguments provided during runtime, such as `use_closest_read_length` or `downgrade_errors`.

Other changes:
- Implement error downgrading
- Add `run_summary` to Illumina view
- QCCheckers and views don't need to be imported individually into `QCReporter` anymore, they will be fetched automatically instead.


checkQC/qc_reporter.py
```diff
@@ -0,0 +1,133 @@
...
+        if any(qc_report.type() == "error" for qc_report in qc_reports):
+            exit_status = 1
+
+        return exit_status, getattr(checkQC.views, config["view"])(
+            checker_configs,
+            qc_data,
+            qc_reports,
+        )
+
+    def _select_configs(
```
matrulda, line 55: I might be missing something obvious here, but why does this class need to re-implement these functions? Could we just use the Config class to find the closest read length and picking the right handler configs? Configs that are used to initialize an object of this class could already be "filtered".

matrulda (commented):
I think it looks great! Just a question regarding the config functions.

checkQC/qc_reporter.py
Aratz: Good question! I felt the `Config` and `ConfigFactory` classes were a bit overkill for something that could in principle be done with a simple dictionary and that only needs to be done once . I wanted to see how it would look like without using such classes.

Maybe we can keep it for now and see what works best when we connect `QCReporter` to the web handler.

matrulda: Yeah, that is a valid point. 

That sounds good :+1:

matrulda (approved):

PR #121, opened 2025-02-20
Added percent_q30 using new QCData and QCReport

nkongenelly:
Adds percent_q30 for BCLConvert using new QCData and QCReport

### DIFF
--- a/checkQC/qc_checkers/percent_q30.py
+++ b/checkQC/qc_checkers/percent_q30.py
@@ -1,5 +1,3 @@
-import numpy as np
-
 from checkQC.handlers.qc_handler import QCErrorFatal, QCErrorWarning
 
 
@@ -25,7 +23,7 @@ def percent_q30(
             "q30": q30,
         }
 
-        read_or_index_text = "index read" if is_index_read else "read"
+        read_or_index_text = "read (I)" if is_index_read else "read"
         msg = ("%Q30 {q30} was too low on lane: {lane} "
                "for {read_or_index_text}: {read}")
 
@@ -40,7 +38,7 @@ def percent_q30(
                 q30 < warning_threshold):
             data["threshold"] = warning_threshold
             return QCErrorWarning(
-                msg.format(**data,read_or_index_text=read_or_index_text),
+                msg.format(**data, read_or_index_text=read_or_index_text),
                 data=data
             )
 
@@ -50,7 +48,7 @@ def percent_q30(
         for lane, lane_data in qc_data.sequencing_metrics.items()
         for read, read_data in lane_data["reads"].items()
         if (qc_report := _qualify_q30(
-            read_data["percent_q30"], lane, read, read_data.get("is_index", False)
+            read_data["percent_q30"], lane, read, read_data.get("is_index")
             )
         )
     ]
--- a/checkQC/qc_checkers/percent_q30.py
+++ b/checkQC/qc_checkers/percent_q30.py
@@ -1,0 +1,58 @@
+import numpy as np
+
+from checkQC.handlers.qc_handler import QCErrorFatal, QCErrorWarning
+
+
+def percent_q30(
+        qc_data,
+        error_threshold,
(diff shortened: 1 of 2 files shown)


checkQC/qc_checkers/percent_q30.py
```diff
@@ -0,0 +1,58 @@
...
+    )
+
+    def _qualify_q30(q30, lane, read, is_index_read):
+        data = {
+            "lane": lane,
+            "read": read,
+            "q30": q30,
+        }
+
+        read_or_index_text = "index read" if is_index_read else "read"
```
Aratz: The index/non index count logic was not re-implemented here (see https://github.com/Molmed/checkQC/blob/master/checkQC/handlers/q30_handler.py#L45-L47). Is that on purpose?

checkQC/qc_checkers/percent_q30.py
```diff
@@ -0,0 +1,58 @@
...
+                data=data
+            )
+
+
+    return [
+        qc_report
+        for lane, lane_data in qc_data.sequencing_metrics.items()
+        for read, read_data in lane_data["reads"].items()
+        if (qc_report := _qualify_q30(
+            read_data["percent_q30"], lane, read, read_data.get("is_index", False)
```
Aratz:
```suggestion
            read_data["percent_q30"], lane, read, read_data["is_index"]
```
I think you can assume this is set by the parser (and if this information  was not found in the interop files, the parser should handle the default value).

tests/qc_checkers/test_percent_q30.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+@pytest.fixture
+def qc_data_and_exp_val():
+    return (
+        namedtuple("QCData", "sequencing_metrics")(
+            {
+                1:{
+                    "reads": {
+                        1: {"percent_q30": 82.},
```
Aratz: Consequently, you should specify `is_index` everywhere here. This also makes the test data more explicit.

tests/qc_checkers/test_percent_q30.py
```diff
@@ -0,0 +1,130 @@
+from collections import namedtuple
+import numpy as np
```
Aratz: This module is not used here, is it?

checkQC/qc_checkers/percent_q30.py
```diff
@@ -0,0 +1,58 @@
+import numpy as np
```
Aratz: This is not used here, is it?

Aratz (commented):
Great! I left some questions and comments for you :)

checkQC/qc_checkers/percent_q30.py
nkongenelly: Oh my! This is so embaracing. I think i abstracted this part from my mind and assumed a lot which should not have been the case. 

I also remember there was a deviation [here](https://canea.snpseq.medsci.uu.se/Case/Case?CaseID=37501)  regarding wrong error_rate in q30 for reads/ index reads.

@Aratz Should this new file implement as per the old one or should it implement according to the suggested new method? (The [task associated with that deviation](https://snpseq.atlassian.net/browse/DATAOPS-999) has not yet been handled but I can handle that too first if needed.)

Aratz: No worries :) yes please implement the new suggested method

nkongenelly:
Thank you for the review. I have now pushed the updated code

checkQC/qc_checkers/percent_q30.py
Aratz: Please fix this according to the suggestion above :)

Using `.get` allows this to fail silently, which might let some bugs slip through.

nkongenelly: Oh, absolutely! I have now  updated the code with this change.

Aratz (approved):
Very good, thanks for addressing my comments :)

PR #122, opened 2025-02-24
Changed how index reads and non-index reads are identified in error m…

nkongenelly:
Ticket: [DATAOPS-999](https://snpseq.atlassian.net/browse/DATAOPS-999)

Updated the terminology in CheckQC so that we refer to all reads and indexes as their “read number”, and then specify if it is an index or not. This would correspond to the way InterOp stats are displayed in MultiQC. Example: A Paried-End run with dual index would have: Read 1, Read 2 (I), Read 3 (I) and Read 4.


checkQC/handlers/error_rate_handler.py
```diff
@@ -47,22 +47,26 @@ def check_qc(self):
...
             if error_rate == 0 and not self.qc_config[self.ALLOW_MISSING_ERROR_RATE]:
-                yield QCErrorFatal("Error rate was found to be 0 on lane: {} for read: {}, this is probably "
+                yield QCErrorFatal("Error rate was found to be 0 on lane: {} for {}: {}, this is probably "
                                    "because there was no PhiX loaded on this lane. If do not use PhiX for your "
                                    "runs you can set 'allow_missing_error_rate' in the config to True, which will "
-                                   "remove the messages in the future.".format(lane_nbr, read))
+                                   "remove the messages in the future.".format(lane_nbr, read_or_index_text, read))
             elif self.error() != self.UNKNOWN and error_rate > self.error():
-                yield QCErrorFatal("Error rate {} was to high on lane: {} for read: {}".format(error_rate,
+                yield QCErrorFatal("Error rate {} was to high on lane: {} for {}: {}".format(error_rate,
```
Aratz: This part would be more readable if it used f-strings and if the lines were formatted to <100 characters

checkQC/handlers/error_rate_handler.py
```diff
@@ -47,22 +47,26 @@ def check_qc(self):
...
                                                                                                read),
                                    ordering=lane_nbr,
                                    data={"lane": lane_nbr, "read": read,
                                          "error_rate": error_rate, "threshold": self.error()})
             elif self.warning() != self.UNKNOWN and error_rate > self.warning():
-                yield QCErrorWarning("Error rate {} was to high on lane: {} for read: {}".format(error_rate,
+                yield QCErrorWarning("Error rate {} was to high on lane: {} for {}: {}".format(error_rate,
                                                                                                  lane_nbr,
+                                                                                                 read_or_index_text,
                                                                                                  read),
```
Aratz: Same comment here

Aratz (commented):
Nice! Could you also apply the same changes to https://github.com/Molmed/checkQC/blob/master/checkQC/qc_checkers/error_rate.py ?

nkongenelly:
Thanks for the review. I have also converted other string formatting to f-string in these 2 files

nkongenelly:
> Nice! Could you also apply the same changes to https://github.com/Molmed/checkQC/blob/master/checkQC/qc_checkers/error_rate.py ?

Oh, I didn't realise I could make this change here in this branch. I will update this then update you when done.

nkongenelly:
@Aratz , I have now completed the changes and the code is ready for your review.

Aratz (approved):
Perfect, thank you for addressing my comments :pray:

matrulda:
Great work, Nelly! 

As of today, error rate is never computed for index reads, i.e. error rate for a index read will be NAN. The error rate is based on comparing the (non-indexed) reads of a known sequence (PhiX) to a reference. This is done by the Illumina software that creates the Interop files. PhiX can be indexed (SNP&SEQ does not use that currently), but in those cases I'm still quite certain that no error rate would be calculated for them since the sequence is so short. However, we do not know if this will change in the future, so I think it still makes sense for the handler to take them into account. However, just reading the code, it would look like error rate is expected for index reads. Do you think this should be communicated somehow? Or would it be better to just skip index reads in the error rate handler / qc checker? To be clear, I think the current implementation works as well, I just wonder from a developer's point of view if this should be addressed.

Also, to be clear, index reads are expected to have a Q30 value, so this concern only applies to the error rate handler/checker.

Sorry for not realizing this sooner! I should have added this information to the ticket.

nkongenelly:
> Great work, Nelly!
> 
> As of today, error rate is never computed for index reads, i.e. error rate for a index read will be NAN. The error rate is based on comparing the (non-indexed) reads of a known sequence (PhiX) to a reference. This is done by the Illumina software that creates the Interop files. PhiX can be indexed (SNP&SEQ does not use that currently), but in those cases I'm still quite certain that no error rate would be calculated for them since the sequence is so short. However, we do not know if this will change in the future, so I think it still makes sense for the handler to take them into account. However, just reading the code, it would look like error rate is expected for index reads. Do you think this should be communicated somehow? Or would it be better to just skip index reads in the error rate handler / qc checker? To be clear, I think the current implementation works as well, I just wonder from a developer's point of view if this should be addressed.
> 
> Also, to be clear, index reads are expected to have a Q30 value, so this concern only applies to the error rate handler/checker.
> 
> Sorry for not realizing this sooner! I should have added this information to the ticket.

Oooh, this is great information. I think maybe we can provide this information as a comment

nkongenelly:
> Great work, Nelly!
> 
> As of today, error rate is never computed for index reads, i.e. error rate for a index read will be NAN. The error rate is based on comparing the (non-indexed) reads of a known sequence (PhiX) to a reference. This is done by the Illumina software that creates the Interop files. PhiX can be indexed (SNP&SEQ does not use that currently), but in those cases I'm still quite certain that no error rate would be calculated for them since the sequence is so short. However, we do not know if this will change in the future, so I think it still makes sense for the handler to take them into account. However, just reading the code, it would look like error rate is expected for index reads. Do you think this should be communicated somehow? Or would it be better to just skip index reads in the error rate handler / qc checker? To be clear, I think the current implementation works as well, I just wonder from a developer's point of view if this should be addressed.
> 
> Also, to be clear, index reads are expected to have a Q30 value, so this concern only applies to the error rate handler/checker.
> 
> Sorry for not realizing this sooner! I should have added this information to the ticket.

@Aratz , what do you think about this? Will having a comment be enough to explain that error rate is not expected for index reads? Or is it better to do away with index reads for error_rate handler?

Aratz:
From what I understand, with the current technology it doesn't make sense to talk about error rate for index reads, so I think the code should reflect that and ignore them too. Otherwise it might create some unnecessary confusion.

nkongenelly:
@matrulda , I have now removed index reads from error rate handler.

checkQC/parsers/interop_parser.py
```diff
@@ -163,8 +163,7 @@ def run(self):
                 self._send_to_subscribers(("error_rate",
                                         {"lane": lane+1, 
                                          "read": read_nbr+1, 
-                                         "error_rate": error_rate,
-                                         "is_index_read":is_index_read}))
```
matrulda: I think it would make sense to keep this, but in the handler skip reads where is_index_read is true. 
That makes it extra clear that the handler is not supposed to handle index reads.

checkQC/qc_checkers/error_rate.py
```diff
@@ -21,41 +21,34 @@ def error_rate(
...
-                    data=data
-                )
+                return QCErrorWarning(msg.format(**data), data=data)
 
     return [
         qc_report
         for lane, lane_data in qc_data.sequencing_metrics.items()
         for read, read_data in lane_data["reads"].items()
-        if (qc_report := _qualify_error(read_data["mean_error_rate"], lane, read, read_data["is_index"]))
+        if (qc_report := _qualify_error(read_data["mean_error_rate"], lane, read))
```
matrulda: Same here, can we make sure that reads where is_index_read = True are skipped?

matrulda (commented):
Looks good, added some comments. Let me know if you think.

nkongenelly:
Thanks @matrulda  for the review. i have now pushed the updated code

checkQC/qc_checkers/error_rate.py
```diff
@@ -49,6 +49,7 @@ def _qualify_error(error, lane, read):
         for lane, lane_data in qc_data.sequencing_metrics.items()
         for read, read_data in lane_data["reads"].items()
         if (qc_report := _qualify_error(read_data["mean_error_rate"], lane, read))
```
Aratz: Actually, for performance reasons, it would be better to write the conditional in the opposite order:
```python
if not read_data["is_index"] and (qc_report := _qualify_error(read_data["mean_error_rate"], lane, read)
```
This way the qc_report is only computed when `is_index` is false, see https://en.wikipedia.org/wiki/Short-circuit_evaluation for details

checkQC/handlers/error_rate_handler.py
```diff
@@ -47,38 +47,41 @@ def check_qc(self):
...
-                        "lane": lane_nbr,
-                        "read": read,
-                        "error_rate": error_rate,
-                        "threshold": self.warning()
-                    }
-                )
-            else:
-                continue
+            # error_rate handler is not supposed to handle index reads.
+            if not is_index_read:
```
matrulda: Instead of doing this, I think it would be cleaner to handle it at the start of the for loop (line 46). Maybe something like this: 
```
for error_dict in filter(lambda x: not x["is_index_read"], self.error_results):
```

checkQC/qc_checkers/error_rate.py
matrulda: Agreed :+1:

matrulda (commented):
Some final comments :)

nkongenelly:
Wooow! I have learnt a lot from this. Thank you @Aratz  and @matrulda for looking into this and for the great tips.

@matrulda , I have now pushed the updated code

matrulda (approved):
Nice, looks great! Thank you!

PR #123, opened 2025-02-28
Implement unidentified index checker

Aratz:
This PR adds a new checker to analyze the top unknown barcodes of a run.

There are many possible causes for a read to appear in this list, so please feel free to suggest any new test case can think of :)


checkQC/handlers/unidentified_index_handler.py
matrulda, line 1: Nice!

checkQC/parsers/illumina.py
```diff
@@ -44,8 +44,10 @@ def from_bclconvert(cls, runfolder_path, parser_config):
             ),
             "top_unknown_barcodes": [
                 {
-                    "index+index2": f"{row['index']}+{row['index2']}",
-                    "count": int(row["# Reads"])
+                    "lane": int(row["Lane"]),
```
matrulda: Is it necessary to store lane information here?

tests/qc_checkers/test_unidentified_index.py
```diff
@@ -0,0 +1,209 @@
...
+        "lane": 1,
+    }
+
+    # OBS: samplesheet rows are reported twice because it matches both indices
+    causes = samplesheet_matcher.check_complement_and_reverse(barcode)
+    assert len(causes) == 6
+    assert any(data[0] == "reverse" for _, data in causes)
+    assert any(data[0] == "complement" for _, data in causes)
+    assert any(data[0] == "reverse complement" for _, data in causes)
+    assert all(data[1]["Sample_ID"].startswith("dual") for _, data in causes)
```
matrulda, line 90: Why do you check this three times?

tests/qc_checkers/test_unidentified_index.py
```diff
@@ -0,0 +1,209 @@
...
+
+
+def test_check_complement_and_reverse(samplesheet_matcher):
+    barcode = {
+        "index": "ACGA",
+        "index2": "AACC",
+        "lane": 1,
+    }
+
+    # OBS: samplesheet rows are reported twice because it matches both indices
```
matrulda: Maybe better to use `NOTE` since I think `OBS` is Swedish? :smile:

checkQC/qc_checkers/unidentified_index.py
```diff
@@ -0,0 +1,193 @@
...
+            )
+
+            data = {"index": index, "lane": lane, "causes": []}
+            msg = (
+                f"Overrepresented unknown barcode \"{data['index']}\" on lane {lane} "
+                f"({significance:.1f}% > {significance_threshold:.1f}%)."
+                + (" This barcode is white-listed." if is_white_listed else "")
+            )
+
+            causes = samplesheet_matcher.list_causes(barcode)
```
matrulda: Related to my other comment about storing lane number on the barcode level, I suppose it would be possible to send the lane as another parameter here and use it downstream? I think the current solution might be cleaner, just something to consider to avoid having the lane number stored in multiple places.

tests/qc_checkers/test_unidentified_index.py
matrulda, line 1: Suggestions on tests to add: 
- Index with whitespace (in samplesheet), I guess that should be be tested by the Illumina parser though. 
- Test indices with different lenghts. For example we do not want undetermined index `CCA`  in lane 2 to match `CCAA` in lane 1. While writing this I become a bit unsure, maybe we would like that actually :thinking: I will sleep on it.

matrulda (commented):
This looks great! Left some comments for you, let me know if you want to address them.

tests/qc_checkers/test_unidentified_index.py
Aratz: Good question! I don't want the tests to depend on the order of the causes here, i.e. one will be `reverse`, one will be `complement` and one will be `reverse complement`, I just want to check that all of them are found in the list of causes

checkQC/qc_checkers/unidentified_index.py
Aratz: Yes, exactly, either that or add it to the barcode dictionary here.

One drawback with adding a `lane` parameter is that it makes it possible have a mismatch between barcode and lane (i.e. send a barcode from lane 1 while the lane parameter is set to 2).

I'm not sure what's better yet, it seems like either way there is a risk for duplication or mismatch. I need to reflect a little bit more on this.

tests/qc_checkers/test_unidentified_index.py
matrulda: Yes agreed, I think that makes sense, but was wondering why you specifically do this assertion three times: 
`assert all(data[1]["Sample_ID"].startswith("dual") for _, data in causes)` 

To me they all look identical, but maybe I'm missing something.

Aratz: Aah, now I see, I got confused because only the first one appears in the code snippet above, yes, they are exactly the same, I think it's a leftover because I use another assertion with `any(...)` before. Good catch, I'll fix!

tests/qc_checkers/test_unidentified_index.py
matrulda: Regarding the second point, I discussed this with Sara and she agrees that sub-string matches might give rise to many false positives and can be confusing.

Aratz: What's the expected behavior for the parser when it encounters indices with whitespaces? Should it keep them or remove them?

As of now it just takes the output from samshee as is, but we could also go through all rows and remove the whitespaces.

checkQC/qc_checkers/unidentified_index.py
Aratz: Actually, maybe a better solution would be to include `barcode` in the `data` dict and pass that as argument to `list_causes` :thinking:

lbeltrame:
I wonder if it makes sense to warn (in case of instruments like the NextSeq family and any which uses the dual fluorophore approach) if poly-G indices are above a (very large) threshold?

Aratz:
Do you mean having a specific, higher threshold for poly-G? At our platform we typically white-list poly-Gs so that they come out as warnings instead of errors. Would that be an acceptable solution?

tests/qc_checkers/test_unidentified_index.py
matrulda: Sorry for the late reply! We want to remove whitespaces.

lbeltrame:
> Do you mean having a specific, higher threshold for poly-G? At our platform we typically white-list poly-Gs so that they come out as warnings instead of errors. Would that be an acceptable solution?

Yeah, that would work perfectly.

matrulda (approved):
Awesome, great work!

PR #124, opened 2025-03-10
Implement undetermined percentage checks

Aratz:
This PR adds a qc_checker to check undetermined percentage and replace the undetermined percentage handler.


checkQC/qc_checkers/undetermined_percentage.py
```diff
@@ -0,0 +1,63 @@
...
+    assert qc_data.sequencing_metrics
+    assert (
+        (error_threshold == "unknown" or warning_threshold == "unknown")
+        or error_threshold > warning_threshold
+    )
+
+    def _qualify_error(lane_data, lane):
+        if lane_data["yield"] == 0:
+            return QCErrorFatal(
+                f"Yield for lane {lane} wa 0. "
```
matrulda:
```suggestion
                f"Yield for lane {lane} was 0. "
```

checkQC/qc_checkers/undetermined_percentage.py
```diff
@@ -0,0 +1,63 @@
...
+        mean_percent_phix_aligned = np.nan_to_num(np.nanmean(
+            [
+                read_data["mean_percent_phix_aligned"]
+                for read_data in lane_data["reads"].values()
+            ]
+        ), nan=0.0)
+
+        percentage_undetermined = (
+            lane_data["yield_undetermined"] / lane_data["yield"] * 100
+            - mean_percent_phix_aligned
```
matrulda: I think reporting `% undetermined` as `% undetermined - % PhiX` will be a bit confusing to the user. It won't match the percentage of reads that was put in the Undetermined fastq for the lane. I think we need to be clear that % PhiX has been subtracted or report it as in the old handler (i.e. write the actual % undetermined, but add data about % phix, computed threshold etc.). 

Either way, I think it would be useful for the user if the mean % PhiX was added to the data dict.

checkQC/qc_checkers/undetermined_percentage.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def _qualify_error(lane_data, lane):
+        if lane_data["yield"] == 0:
+            return QCErrorFatal(
+                f"Yield for lane {lane} wa 0. "
+                "No undetermined percentage could be computed",
+                data={"lane": lane, "percentage_undetermined": None},
+            )
+
+        mean_percent_phix_aligned = np.nan_to_num(np.nanmean(
```
matrulda, line 28: Nice use of nan_to_num!

matrulda (commented):
Looks great! I left some comments for you.

checkQC/qc_checkers/undetermined_percentage.py
Aratz: Thanks :)

matrulda (approved):
Awesome, looks great! Thanks for addressing my comments.

PR #125, opened 2025-03-27
Fix reads per samples

Aratz:
A subtle bug found its way in the reads per sample checker: because of this extra comma on line 57, `qc_report` was packed into a tuple and the `if`-statement sometimes looked like `if (None,)`, which surprisingly enough always return `True`.

This PR removes this rogue comma and adds assertions to make sure qc_checkers always return valid reports.

### DIFF
--- a/tests/qc_checkers/test_reads_per_sample.py
+++ b/tests/qc_checkers/test_reads_per_sample.py
@@ -98,9 +98,8 @@ def test_reads_per_sample_unknown_threshold(qc_data_and_exp_val):
             "lane": 1,
             "number_of_samples": 2,
             "sample_id": "Sample_B",
-            "sample_reads": 20,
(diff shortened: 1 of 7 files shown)


tests/qc_checkers/test_reads_per_sample.py
```diff
@@ -69,8 +69,9 @@ def test_reads_per_sample(qc_data_and_exp_val):
         warning_threshold=60,
     )
 
-    assert len(qc_reports) == 4
-    for qc_report in filter(None, qc_reports):
```
matrulda, line 73: Ah yes, now I remember, I assumed that it was expected that QC reports could be None :facepalm:

matrulda (approved):
Looks good to me!

PR #126, opened 2025-03-31
Small fixes to new parsers and checkers

Aratz:
While testing the new CheckQC manually, I encountered a few bugs, that this PR addresses:
- When parsing a yaml file, the read length will be parsed as integer if it's a single number, or as a string if it is a range. We need to make sure it is read as a string when matching it with a regex
- Illumina parser:
  - In case of pair reads, the run type recognizer will return both read lengths (eg. `151-151`)
  - On basespace, I found samplesheets with "index" and some with "Index" in the column names. Thus we need to support both.
- One exception message was found to be missing a space


checkQC/parsers/illumina.py
```diff
@@ -105,25 +107,53 @@ def _read_interop_summary(runfolder_path):
...
+
+    for i, row in enumerate(samplesheet):
+        samplesheet[i] = {
+            key.lower(): value
+            for key, value in row.items()
+        }
+
+    for row in samplesheet:
+        row["index"] = row["index"].replace(" ", "")
+        row["index2"] = row["index2"].replace(" ", "")
```
nkongenelly, line 157: From the description of the bugs fixed here, it is mentioned that we want to find either "index" or "Index" in the column names. I might be missing something here as i think i see only getting the "index".

Might the implementation to get either "index" or "Index" have been implemented somewhere else?

nkongenelly (commented):
Great work.It looks great. Just left a clarification question  but otherwise it looks great.

checkQC/parsers/illumina.py
Aratz: It is just a few lines above ;)
https://github.com/Molmed/checkQC/blob/70954d4f91a0f8774af019f4404cb6b6e2fa4cb7/checkQC/parsers/illumina.py#L151

but maybe I can clarify what I was thinking when I said we should support both:

We just want to be able to parse both types of samplesheets: the ones with `index` and the ones with `Index`, but we don't want to have to try both `index` and `Index` whenever we try to access a row of the samplesheet, that would be a quite cumbersome. Instead we want to make sure that everything is stored with lowercase characters internally. So as soon as we read the data from the samplesheet we want to cast the keys to lowercase.

nkongenelly: Ooh, okay. Tack.

nkongenelly (approved):

PR #127, opened 2025-03-31
Improve Illumina view

Aratz:
When showing the current view to @KarNair, it came up that the view would be clearer if error messages were grouped by error type. This PR addresses that.

This PR also implements a basic view that just outputs the QC reports as a list without any further sorting. This can be used e.g. when analyzing the data from a non-illumina instrument.


checkQC/views/illumina.py
```diff
@@ -8,32 +8,34 @@ def illumina_view(checker_configs, qc_data, qc_reports):
...
                 key=lambda report: report.data.get("lane")
             )
         },
-        "other reports": [
-            str(report)
-            for report in qc_reports
-            if "lane" not in report.data
-        ],
+        "other reports": {
+            qc_checker: non_lane_reports
```
KarNair: Is there a way to find which of the qc-checkers are non-lane? This is more for review purposes

Aratz: All those that does not contain the key `lane` in the `data` dictionary they output, but currently all checkers are specific to a lane.

This section is here in case we add checkers later that do not specify a lane.

KarNair: In that case, if we do not have any specific non-lane checkers in mind, maybe we should not have this section at all? Atleast as of now, it feels like non-lane checkers are a bit hypothetical

checkQC/parsers/illumina.py
```diff
@@ -22,10 +22,7 @@ def from_bclconvert(cls, runfolder_path, parser_config):
         / parser_config["reports_location"]
         / "Top_Unknown_Barcodes.csv"
     )
```
KarNair, line 24: Well, because Github doesn't let you comment on lines not affected by the PR (GitLab Supremacy), you'll have to bear with me on this one.

This is nitpicky, but I feel like the file paths in lines 14-18, and line 21-24 could be indented/formatted better maybe? It took me a while before realising that these are filepaths. Right off the bat, it looked like a bunch of variable names indented in a weird way without commas.

KarNair (commented):
I have a few comments. Other than that it looks good! The illumina view looks perfect!

checkQC/parsers/illumina.py
Aratz: The main reasons it's implemented like this are because using slashes to combine path is the standard way with pathlib (https://docs.python.org/3/library/pathlib.html#operators), and using line breaks before operators is the recommended way according to PEP8 (https://peps.python.org/pep-0008/#should-a-line-break-before-or-after-a-binary-operator). 

However, if you have any suggestions on how to make it more readable and respect these constraints, I'm all in :)

checkQC/views/illumina.py
Aratz: Hmm, I see, one problem I can think about with removing it is that, if we implement such lane-free checker, it might be difficult to understand why it does not show up in the view.

I'll add some checks in the view to make it explicit that it expects the `data` dict to have a `lane` field :+1:

checkQC/parsers/illumina.py
KarNair: I stand corrected! I was not aware of the line break convention. This makes more sense. Thank you!

KarNair:
This looks good!

KarNair (approved):

PR #128, opened 2025-04-02
Connect `QCData` to cli and web app

Aratz:
This PR connects the app's entry points to the new CheckQC implementation, namely the QCData and QCReporter classes.

It introduces a new parameter `--demultiplexer`, which specifies which demultiplexer was used to generate the fastqfiles and indicate to CheckQC how the Interop files should be parsed.

For backward compatibility, we maintain the old behavior when analyzing data demultiplexed  with `bcl2fastq`. The new implementation should be for all other demultiplexers.

Eventually, we should also use the new implementation for `bcl2fastq` too and remove the old code.


checkQC/app.py
```diff
@@ -37,18 +70,92 @@ def start(config, json, downgrade_errors, use_closest_read_length, runfolder):
...
+    use_closest_read_length,
+    demultiplexer,
+):
+    runfolder_path = Path(runfolder_path)
+    assert runfolder_path.is_dir()
+
+    config = ConfigFactory.from_config_path(config)._config
+
+    try:
+        qc_data_constructor = getattr(QCData, f"from_{demultiplexer}")
```
nkongenelly: "Would it be a good idea to convert {demultiplexer} to lowercase, just in case someone writes it as 'BCLConvert' or in another variation? Or should we be strict and only allow lowercase without conversion?"

tests/test_integration_cli.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+@pytest.fixture
+def runfolder_data():
+    runfolder_path = str(Path(__file__).parent / "resources/bclconvert/200624_A00834_0183_BHMTFYTINY/")
+    expected_data = {
+        "lane reports": {
+            "1": {
+                "error_rate": [
+                    "Fatal QC error: Error rate is nan on lane 1 for read 1. This may be because no PhiX was loaded on this lane. Use \"allow_missing_error_rate: true\" to disable this error message."
```
nkongenelly: These lines seem a bit long and might be better splitting into multiple lines.

nkongenelly (commented):
This looks great! I've learned a few things from this (e.g `getattr()` part) and have left some minor comments. 

Also, will you be increasing the CheckQC version in this PR, or are there more sub-tickets to address before bumping the version?

Otherwise it looks great!

checkQC/app.py
```diff
@@ -22,13 +23,45 @@
...
+    help="Downgrade errors to warnings for a specific handler, can be used multiple times",
+)
+@click.option(
+    "--use-closest-read-length",
+    is_flag=True, default=False,
+    help="Use the closest read length if the read length used isn't specified in the config",
+)
+@click.option(
+    "--demultiplexer",
+    type=str,
```
matrulda: Non-critical comment: Using the Choice type could be an option here: https://click.palletsprojects.com/en/stable/options/#choice-options

checkQC/app.py
matrulda: Non-critical comment: This could be solved by using the Choice type with `case_sensitive=False` (see previous comment).

tests/resources/200624_A00834_0183_BHMTFYTINY/SampleSheet.csv
matrulda, line 1: To avoid potential duplicate data, it would be good to address this: 
Is this runfolder the same as the one in the bclconvert directory? If yes, can we remove that one or make it into a symlink to `tests/resources/200624_A00834_0183_BHMTFYTINY`?

checkQC/app.py
```diff
@@ -37,18 +70,92 @@ def start(config, json, downgrade_errors, use_closest_read_length, runfolder):
...
+        log.info(f"Runfolder is: {runfolder}")
+
+        exit_status, reports = run_new_checkqc(
+            config,
+            runfolder,
+            downgrade_errors,
+            use_closest_read_length,
+            demultiplexer,
+        )
+
```
matrulda, line 111: Critical comment: Something I am missing in the CLI is outputting the results (if `json` is not used). In the old implementation is it done as part of compiling the reports: https://github.com/Molmed/checkQC/blob/4be8e67e1d185f85a785eaf145d0dae5abad4041/checkQC/handlers/qc_handler.py#L242

To behave the same way in the new implementation, we should loop trough the reports and log QCErrorFatal as errors and QCErrorWarning as warnings. Personally I think it might be better to log them all as info messages, since error log messages should indicate that something went wrong in the application. Let me know what you think.

tests/test_integration_cli.py
```diff
@@ -0,0 +1,118 @@
...
+                ],
+                "reads_per_sample": [
+                    "Fatal QC error: Number of reads for sample Sample_14574-Qiagen-IndexSet1-SP-Lane1 on lane 1 were too low: 0.00992 M (threshold: 121.875 M)",
```
matrulda: Somewhat critical comment:
Something I realized that I miss in the views is dumping the data dict a long side the message. I think some users might be parsing that. 
```
"reads_per_sample": [
                    {"message": "Fatal QC error: Number of reads for sample Sample_14574-Qiagen-IndexSet1-SP-Lane1 on lane 1 were too low: 0.00992 M (threshold: 121.875 M)",
                     "data": {
            		"lane": <lane>,
            	        "number_of_samples": <number_of_samples>,
            	        "sample_id": <sample_id>,
                        "sample_reads": <sample_reads>,
                        "qc_checker": "reads_per_sample",},
                    {"message": "Fatal QC error: Number of reads for sample Sample_14575-Qiagen-IndexSet1-SP-Lane1 on lane 1 were too low: 0.00856 M (threshold: 121.875 M))",
                     "data": {
            		"lane": <lane>,
            	        "number_of_samples": <number_of_samples>,
            	        "sample_id": <sample_id>,
                        "sample_reads": <sample_reads>,
                        "qc_checker": "reads_per_sample",},
                ],
[...]
```

It makes the output a bit less readable, so we could make dumping the data dict an option?

matrulda (changes requested):
Great work! Left some comments for you!

checkQC/app.py
Aratz: Great ideas!

tests/resources/200624_A00834_0183_BHMTFYTINY/SampleSheet.csv
Aratz: Good point. I think a more general issue is that the `resources` directory is a bit messy. I'll try to restructure it a bit, maybe something like: one folder per demultiplexer, with the relevant runfolders inside, and then a `monitored_dir` with links to the relevant runfolders

checkQC/app.py
Aratz: Ah, I had missed that. I agree that it would be nice to log all the reports.

However, I don't think the logs should be used instead of the standard output. I think it should be more for traceability.

Maybe we should refine what CheckQC outputs: 
- by default, it prints the report in a human readable way (mostly the report messages, grouped as defined in the view).
- if `--json` is used, it outputs the json generated for the view instead.
- In any case, whenever a report is generated, it is logged at the info level (but not grouped according to a view, this is just to get a record of what was generated)

Any thoughts @matrulda?

tests/test_integration_cli.py
Aratz: Good point, I'll address that together with the previous comment.

checkQC/app.py
matrulda: That sounds great! :+1:

tests/test_integration_cli.py
Aratz: This was harder to address than I expected :sweat_smile: I wanted to avoid splitting it over multiple lines because I figured it would make it a bit too cumbersome to update if we change the message in the future, so I decided to move all of this in a separate file that I load at the beginning of the test :+1:

Aratz:
@matrulda Took me a while, but I think I finally landed on something I'm finally happy with. In the end I've decided it is best if everything related to output is moved to the views.

checkQC/app.py
```diff
@@ -70,6 +81,13 @@ def start(
     # -----------------------------------
     # This is the application entry point
     # -----------------------------------
+
+    if json_mode:
+        warnings.warn(
```
matrulda: Nice! This made me look into https://stackoverflow.com/questions/9595009/warnings-warn-vs-logging-warning . Thanks for that knowledge gift! 

One thing though, I tried triggering this and the warning message gets a bit cut off (see `warnings.warn(`): 
```
$ checkqc tests/resources/bclconvert/200624_A00834_0183_BHMTFYTINY --use-closest-read-length --demultiplexer bclconvert --config bclconvert_config.yaml --json 
/home/matilda/workspace/checkQC/checkQC/app.py:86: DeprecationWarning: `--json` is being deprecated in favor of custom views and only works when the demultiplexer is bcl2fastq.
  warnings.warn(
INFO     ------------------------
INFO     Starting checkQC (4.0.6)
INFO     ------------------------

```

I tried updating line 86 (in app.py) so that the whole warnings.warn command in printed on one line, and it looks like it helps: 
```
$ checkqc tests/resources/bclconvert/200624_A00834_0183_BHMTFYTINY --use-closest-read-length --demultiplexer bclconvert --config bclconvert_config.yaml --json 
/home/matilda/workspace/checkQC/checkQC/app.py:86: DeprecationWarning: `--json` is being deprecated in favor of custom views and only works when the demultiplexer is bcl2fastq.
  warnings.warn("`--json` is being deprecated in favor of custom views and only works when the demultiplexer is bcl2fastq.", DeprecationWarning)
INFO     ------------------------
INFO     Starting checkQC (4.0.6)
INFO     ------------------------

```
So this must be some bug in warnings. Also, unclear why the command is printed at all? The first line contains all info needed. Perhaps it can be omitted by using another simplefilter? 

Anyho, this is a very minor issue, since the Deprecation warning information is visible. Just wanted to bring it up.

README.md
```diff
@@ -209,6 +209,32 @@ INFO     Finished with fatal qc errors and will exit with non-zero exit status.
...
+NB: for these demultiplexers, the output is defined by the view classes. So
+far the following views are available:
+
+- `basic_view`: outputs a json string containing all reports in a list, as well
+  as another dict summarizing the run parameters. Only the report's messages
+  are returned
+- `illumina_data_view`: outputs a json string where reports are gathered by lane
+and by type of report. Run summary is also attached. Reports include both the
+message string and the data dictionary.
+- `illumina_short_view`: outputs a yaml string gathering all reports by lane
```
matrulda, line 231: Run summary is included here as well. Also, is it worth mentioning that only the messages are included here?

README.md
```diff
@@ -209,6 +209,32 @@ INFO     Finished with fatal qc errors and will exit with non-zero exit status.
...
+
+```
+checkqc --demultiplexer bclconvert <RUNFOLDER>
+```
+
+NB: for these demultiplexers, the output is defined by the view classes. So
+far the following views are available:
+
+- `basic_view`: outputs a json string containing all reports in a list, as well
+  as another dict summarizing the run parameters. Only the report's messages
```
matrulda: Maybe overkill, but to reduce the risk of confusion, call this run summary as well. Because it is the same right?

README.md
```diff
@@ -209,6 +209,32 @@ INFO     Finished with fatal qc errors and will exit with non-zero exit status.
...
+  are returned
+- `illumina_data_view`: outputs a json string where reports are gathered by lane
+and by type of report. Run summary is also attached. Reports include both the
+message string and the data dictionary.
+- `illumina_short_view`: outputs a yaml string gathering all reports by lane
+and report type. This is suitable for cases where CheckQC's output is monitored
+by a human operator.
+
+Views can be selected in the config file.
```
matrulda, line 236: Based on how the default config looks, I would, as a regular user, assume that the view setting has to be set at the global "default level". It might be worth mentioning that it can be set on read length level. 

Also, it would be sweet if it could be set on run type level, something like:
```
novaseq_SP:
  51:
    handlers:
      - name: ClusterPFHandler
        warning: 325
        error: unknown
      - name: Q30Handler
        warning: 90
        error: unknown
      - name: ErrorRateHandler
        allow_missing_error_rate: False
        warning: 1
        error: unknown
      - name: ReadsPerSampleHandler
        warning: unknown
        error: 243.75
  101:
    handlers:
      - name: ClusterPFHandler
        warning: 325
        error: unknown
      - name: Q30Handler
        warning: 85
        error: unknown
      - name: ErrorRateHandler
        allow_missing_error_rate: False
        warning: 2
        error: unknown
      - name: ReadsPerSampleHandler
        warning: unknown
        error: 243.75
  151:
    handlers:
      - name: ClusterPFHandler
        warning: 325
        error: unknown
      - name: Q30Handler
        warning: 85
        error: unknown
      - name: ErrorRateHandler
        allow_missing_error_rate: False
        warning: 5
        error: unknown
      - name: ReadsPerSampleHandler
        warning: unknown
        error: 243.75
  251:
    handlers:
      - name: ClusterPFHandler
        warning: 325
        error: unknown
      - name: Q30Handler
        warning: 75
        error: unknown
      - name: ErrorRateHandler
        allow_missing_error_rate: False
        warning: 5
        error: unknown
      - name: ReadsPerSampleHandler
        warning: unknown
        error: 243.75
  view: illumina_short_view
```
I think it is out of scope for this PR though, but if you think it is a good idea I can create a ticket.

README.md
```diff
@@ -209,6 +209,32 @@ INFO     Finished with fatal qc errors and will exit with non-zero exit status.
 }
 ```
 
+
+Running CheckQC (other demultiplexers)
```
matrulda, line 213: Just a side note is that this information is missing from the docs (which is not building anyway https://github.com/Molmed/checkQC/issues/133). I think we need to decide if/how we want to use the docs, so no action needed.

tests/resources/200624_A00834_0183_BHMTFYTINY/SampleSheet.csv
matrulda: Great, it is so nice and organized now :star_struck:

checkQC/views/illumina.py
```diff
@@ -1,16 +1,41 @@
...
+    Group reports by lane and checker type.
     """
     assert all("lane" in report.data for report in qc_reports)
 
     return {
-        "lane reports": {
+        "lane_reports": {
             lane: {
-                qc_checker: [str(report) for report in reports]
+                qc_checker: [format_report(report) for report in reports]
```
matrulda, line 38: Very neat solution :clap:

matrulda (commented):
Great work, I really like how you solved this with the views! 

Left some minor comments, let me know if you want to address them or not.

checkQC/app.py
Aratz: I found this solution but I don't find it very pretty, maybe it is better to keep the default behavior?
https://stackoverflow.com/questions/2187269/print-only-the-message-on-warnings

matrulda: @Aratz Yeah I agree, it adds unnecessary complexity. If this can't be fixed with one line, I think it is better to keep the default behavior.

README.md
Aratz: Good point! I'll fix the default config to make it clear the view can be specified at the read level, and for the run type level I agree it can be a ticket. Maybe we could also consider if it would be advantageous to define a new config format at the same time .

README.md
Aratz: Good point, will fix!

checkQC/views/illumina.py
Aratz: Thanks :blush:

Aratz:
Thanks for the review! Hopefully it should be good to go now

README.md
matrulda: yeah, could be a good opportunity to define a new config format that we over time can transition fully to. I created a ticket: https://snpseq.atlassian.net/browse/DATAOPS-1164

matrulda (approved):
Great, thank you!

PR #134, opened 2025-08-27
Dataops 1178 update checkqc for projman

nkongenelly:
[Jira ticket](https://snpseq.atlassian.net/browse/DATAOPS-1178)

This version updates the sequencing_metrics returned by the illumina parser to contain all metrics needed by projman_filler to save flowcell_lane and sample metrics


checkQC/parsers/illumina.py
```diff
@@ -13,23 +13,33 @@ def from_bclconvert(cls, runfolder_path, parser_config):
...
+        / "Demultiplex_Stats.csv"
+    )
     samplesheet = _read_samplesheet(runfolder_path)
 
     instrument, read_length = _read_run_metadata(runfolder_path)
 
     sequencing_metrics = {
         lane + 1: {
             "total_cluster_pf": summary.at(0).at(lane).reads_pf(),
+            "pf_clusters": summary.at(0).at(lane).cluster_count_pf().mean(),
```
matrulda: How does this differ from `total_cluster_pf`? In my mind it would be the same? Though Reads might be times 2 when sequencing paired-end. 

I think we might have verified this already, but just to make sure: was is necessary to use the mean function for this value and the ones below? I asking since the reads_pf could be fetch without using the mean function.

checkQC/parsers/illumina.py
```diff
@@ -69,6 +79,36 @@ def from_bclconvert(cls, runfolder_path, parser_config):
                         sample_summary := index_summary.at(lane).at(sample_no)
                     ).sample_id(),
                     "cluster_count": sample_summary.cluster_count(),
+                    "percent_of_lane":next(
```
matrulda:
```suggestion
                    "percent_of_lane": next(
```

checkQC/parsers/illumina.py
```diff
@@ -69,6 +79,36 @@ def from_bclconvert(cls, runfolder_path, parser_config):
...
+                    "mean_q30": next(
+                        float(row["Mean Quality Score (PF)"])
+                        for row in quality_metrics
+                        if (
+                            row["Lane"] == str(lane + 1)
+                            and row["SampleID"] == sample_summary.sample_id()
+                        )
+                    ),
+                    "percent_q30": next(
+                        float(row["% Q30"])
```
matrulda: I think this should be times 100 as well? The Test file specified this value as e.g. 0.96

tests/test_qc_data.py
```diff
@@ -35,33 +35,41 @@ def bclconvert_runfolder():
...
                     "index2": "CTACAAGATA",
                     "sample_project": "CD-5678",
+                    "description": "LIBRARY_NAME:test",
                 },
             ],
         },
         "expected_sequencing_metrics": {
             1: {
                 "total_cluster_pf": 532_464_327,
+                "pf_clusters": 3_413_232.5,
```
matrulda: Hmm I think we should take another look at if these numbers make sense. It differs so much from the total_clusters_pf.

setup.py
```diff
@@ -11,14 +11,19 @@
     author_email='johan.dahlberg@medsci.uu.se',
     url="https://www.github.com/Molmed/checkQC",
     download_url='https://github.com/Molmed/checkQC/archive/{}.tar.gz'.format(__version__),
-    python_requires='>3.10, <3.11',
```
matrulda, line 14: It would be nice to update the GHA workflow to check if tests works for 3.11, 3.12 and 3.13 as well :) 
This is not necessary though

matrulda (commented):
Looks good! Left some comments for you.

checkQC/parsers/illumina.py
nkongenelly: Oh, i tried to check if pf_clusters might just be total_clusters/2 but the differences in values are quite large e.g when total_cluster_pf =532464327, the pf_clusters value is 3413232.5 ( total_cluster_pf here is about 156 times more than pf_clusters value)
And the pf_clusters value in this instance is what matches what is needed to be saved  to the FlowcellLane Table (comparing with the values found when using bcl2fastq)

tests/test_qc_data.py
nkongenelly: You're absolutely right

checkQC/parsers/illumina.py
nkongenelly: Yes, true.

setup.py
nkongenelly: I have confirmed that GHA workflow here works with python 3.11, 3.12 and 3.13 too and now left the GHA python version as python v3.13

nkongenelly:
Thanks for the review. I have pushed the updated code and maybe we can have a meeting to discuss the values again since as also mentioned that total_clusters_pf is much bigger than pf_clusters.

.github/workflows/unit_tests.yml
```diff
@@ -8,10 +8,10 @@ jobs:
     steps:
     - uses: actions/checkout@v4
     
-    - name: Set up Python 3.10
+    - name: Set up Python 3.13
       uses: actions/setup-python@v4
       with:
-        python-version: '3.10'
+        python-version: '3.13'
```
matrulda: I think it could be nice to use a test matrix so that we make sure to test all supported versions in the GHA workflow: https://stackoverflow.com/a/61428673

matrulda:
> Thanks for the review. I have pushed the updated code and maybe we can have a meeting to discuss the values again since as also mentioned that total_clusters_pf is much bigger than pf_clusters.

Great work! I left a final comment for you and we definitely should look a meeting to look at the values.

.github/workflows/unit_tests.yml
nkongenelly: Oh wow!, this is interesting, thank you for this and I have now pushed the updated code with this change.

nkongenelly:
> > Thanks for the review. I have pushed the updated code and maybe we can have a meeting to discuss the values again since as also mentioned that total_clusters_pf is much bigger than pf_clusters.
> 
> Great work! I left a final comment for you and we definitely should look a meeting to look at the values.

I have booked a meeting and sent you an invite email for this

.github/workflows/unit_tests.yml
```diff
@@ -5,13 +5,17 @@ on: [push, pull_request]
 jobs:
   build:
     runs-on: ubuntu-22.04
+    strategy:
+      matrix:
+        python-version: ['3.10', '3.11', '3.12','3.13']
+    name: Set up Python ${{ matrix.python-version }
```
matrulda:
```suggestion
    name: Set up Python ${{ matrix.python-version }}
```

.github/workflows/unit_tests.yml
```diff
@@ -5,13 +5,17 @@ on: [push, pull_request]
...
     steps:
     - uses: actions/checkout@v4
     
-    - name: Set up Python 3.10
+    - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v4
       with:
-        python-version: '3.10'
+        python-version: ${{ matrix.python-version }
```
matrulda:
```suggestion
        python-version: ${{ matrix.python-version }}
```

matrulda (commented):
Nice, but I think you forgot some curly brackets

nkongenelly:
> Nice, but I think you forgot some curly brackets

Oh yes! Sorry for that. I think i was in my branch that i didn't see the unit tests fail. I have now updated this.

nkongenelly:
Hej @matrulda , I have now pushed the updated code after we finalized samplesheet v2 structure.

tests/test_qc_data.py
```diff
@@ -35,33 +35,40 @@ def bclconvert_runfolder():
...
                     "sample_project": "CD-5678",
+                     "custom_description": "LIBRARY_NAME:test",
                 },
             ],
         },
         "expected_sequencing_metrics": {
             1: {
-                "total_cluster_pf": 532_464_327,
+                "total_reads_pf": 532_464_327,
+                "total_reads": 638337024,
```
matrulda: I think it would be nice to use the "underscore syntax" here as well.

matrulda (commented):
Great, looks good! Just a minor comment.

tests/test_qc_data.py
nkongenelly: I agree, thanks

nkongenelly:
Hej @matrulda , a kind reminder on this PR so that we can use the master branch when this is merged in projman_filler to install checkqc

nkongenelly:
I have just pushed some changes to enable projman_filler to;

-  get flowcell_id (i.e from checkqc interop run_info ) and 

- also moved some test data inside the checkQC modue so that it can be used by projman_filler tests too ( i tried to include checkqc tests/ test_qc_data.py to be among the files avaliable to import from projman_filler by editing local manifest.in file and setup.py and also using the #egg in the install command but it seemed not to work or i might have implemented them wrongly.....that's why i ended up moving the test_data inside the module folder) ... here is a [stackoverflow link ](https://stackoverflow.com/questions/32390705/should-i-include-tests-and-pyc-files-when-building-package-with-setuptools#:~:text=The%20discussion%20is%20getting%20long%2C%20so%20I%27ll%20just%20write%20up%20my%20thoughts%20here.) where i got the basic information from

nkongenelly:
I also tried to move the qc_data_utils file outside the checkQC folder, and adding it is setup.py (under the package_data list) but i was unsuccessful in importing it in the projman_filler side

And i have now added the OverrideCycles in the bclconvert samplesheet (i had forgotten about it in my previous commit)

checkQC/qc_data_utils.py
```diff
@@ -0,0 +1,195 @@
+import numpy as np
+
+
+def bclconvert_test_runfolder(qc_data):
```
matrulda: The only issue I have with this function is that it looks like it is a general utility tool, but it expects qc_data to be based on "200624_A00834_0183_BHMTFYTINY" to match the "expected_*" values. 
Could we move this part to the function as well? 
```
qc_data = QCData.from_bclconvert(
        Path(__file__).parent / "resources/bclconvert/200624_A00834_0183_BHMTFYTINY",
        parser_config,
    )
```

I guess we won't be able to parse the CheckQC test resource when using this function in projman_filler, but in that case maybe the input to this function could be the path to 200624_A00834_0183_BHMTFYTINY? It should state clearly in the doc string that this runfolder is expected as input. 
Let me know what you think.

matrulda (commented):
Great, I had one question.

checkQC/qc_data_utils.py
nkongenelly: Yes, you're right, this function is only tied to 200624_A00834_0183_BHMTFYTINY.  thought of using qc_data to make the function dynamic but then realised that would be comparing 2 qc_data that are generated a fresh each time which may always match.

Yes, i think the function can take runfolder as input

nkongenelly: Or maybe check if the flowcell_id for the TINY runfolder is "HMTFYDRXX"  because the bclconvert runfolder in projman_filler has a slightly different name i.e 200624_A00834_0184_BHMTFYTINY.

nkongenelly: I have now updated the code passing the runfolder Path and checking for the flowcell_id. I have also updated projman_filler to pass the runfolder Path

But this is change is also open for discussion :smiley:

checkQC/qc_data_utils.py
```diff
@@ -0,0 +1,199 @@
+import numpy as np
+from checkQC.parsers.illumina import _read_interop_summary
+
+
+def bclconvert_test_runfolder(qc_data, runfolder_path):
+    _, _, run_info = _read_interop_summary(runfolder_path)
+    flowcell_id = run_info.flowcell_id()
+    if "HMTFYDRXX" in flowcell_id:
```
matrulda, line 8: Great! Just one minor thing, I think it would be nice to throw an exception if the flowcell ID does not match. Explaining that the ouytput of this funtion is adapated for a specific run

nkongenelly: Oh yes, thanks

nkongenelly:
Thanks for the reviews, I have now updated the code.

checkQC/qc_data_utils.py
```diff
@@ -196,4 +196,8 @@ def bclconvert_test_runfolder(qc_data, runfolder_path):
                 },
             },
         }
+    else:
+        raise Exception("Excpected flowcell_id value as 'HMTFYDRXX' only for "
+                        f"this fuction but got {flowcell_id}"
```
matrulda: Oh, there is a typo in the exception. Also, I thought the exception could be rephrased to something like this. 

```suggestion
        raise Exception("This function is only compatible with the run with flowcell_id: 'HMTFYDRXX', "
                                  f"the supplied runfolder has flowcell_id: {flowcell_id}"
```

matrulda:
Just one super minor thing, then we can merge this! :)

nkongenelly:
Thanks for the reviews. I have now updated the code

matrulda (approved):

PR #135, opened 2026-09-15
Add support for 5B FC

sarek928:
Updated the config with values for Novaseq X Plus 5B FC. Changed error rate value for all Novaseq X Plus flowcells and reads lengths to 2.5.

Risk analysis: if the incorrect value has been added to the config, samples with poor quality might pass through QC unnoticed. There should be no issues with running the software.

Validation, it would be good to run 5B testdata + and already processed run with the config updates to make sure that everything works as expected.


matrulda (approved):
Looks good to me! I compared the values with the specs on [Illumina's website](https://www.illumina.com/systems/sequencing-platforms/novaseq-x-plus/specifications.html) and our guidelines in INS-00123.

For validation, I checked BaseSpace, but it doesn't look like Illumina has published any 5B runs yet. I did, however, run this version of CheckQC against a production 25B run, and it completed successfully. So while I couldn't verify the 5B values directly, at least these changes don't seem to have introduced any config syntax issues or similar problems.

PR #136, opened 2026-09-16
Bump version to v4.1.1

matrulda:
Description: Before merging https://github.com/Molmed/checkQC/pull/135 I think we should make a proper release of v4.1.1 (seems like we forgot to do that after it  was verified 6 months ago :sweat_smile:). I will also upload it to pypi.

For the record: I used to do version updates like this directly on master, but I have now introduced branch protection. We could add bypass rules for maintainers, but I will avoid that for now. 

Risk analysis: No risks identified. No changes introduced except the version number

Validation procedure: The release candidate has already been verified.


nkongenelly (approved):
Looks good.

